PR #7, opened 2019-08-21
Add composer.json & propose Packagist release

kadamwhite:
Hello, and thank you for writing a useful plugin!

I am trying to use this plugin within one part of a highly nested, Composer-managed build process for a WP site, and I'm currently unable to do so because I cannot depend on this package via Composer — only the root package can add repository metadata to the collection of known packages, so neither a direct GitHub reference nor the excellent [WPackagist](https://wpackagist.org) service work in a highly nested dependency. We would therefore ideally be able to pull this plugin down from [Packagist](https://packagist.org) directly.

This PR introduces a `composer.json` in the hopes that you might be able to merge this and publish the plugin directly to Packagist.org.

Fingers crossed, and thanks again!


darylldoyle:
Hi @kadamwhite,

Thanks for the PR 🙂 

I see no reason I can't do that for you. I'll get this merged and added to Packagist this evening. I'll update the PR on your side when this is done!

Cheers!

kadamwhite:
Thank you very much!

PR #8, opened 2019-10-31
Rename licence.txt to license.txt

ge0ffray:


jeffpaul:
This naming format more closely matches our 10up standard.

PR #11, opened 2020-05-06
Only attempt to parse dimensions if numeric

joehoyle:
Currently images uploaded with `width="100%"` attributes will get parsed as `100`, so the dimentions are incorrect. Given the following image:

```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:serif="http://www.serif.com/" width="100%" height="100%" viewBox="0 0 500 250"
```

Will result in an image which is `100 x 100`, but we actually want it to be parsed as `500 x 250`.

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -426,7 +426,7 @@ if ( ! class_exists( 'safe_svg' ) ) {
             $height = 0;
             if ( $svg ) {
                 $attributes = $svg->attributes();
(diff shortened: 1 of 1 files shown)


darylldoyle:
Hi @joehoyle 👋 

Good catch there, not too sure how I managed to miss that happening. Thanks for the PR, it's much appreciated! 

I'll get a new release, including this, tagged this evening 🙂

PR #12, opened 2021-02-04
PHP 8.0: libxml_disable_entity_loader function is deprecated

amdd-tim:
commit checks PHP and LIBXML_VERSION before calling libxml_disable_entity_loader.

Feel free to noodle how you see fit. I was getting a fatal error on my Kinsta install, so I had to make the change anyway. See https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation for further insight.

### DIFF
--- a/lib/vendor/enshrined/svg-sanitize/src/Sanitizer.php
+++ b/lib/vendor/enshrined/svg-sanitize/src/Sanitizer.php
@@ -245,8 +245,11 @@ class Sanitizer
      */
     protected function setUpBefore()
     {
-        // Turn off the entity loader
-        $this->xmlLoaderValue = libxml_disable_entity_loader(true);
+        
+        if (\PHP_VERSION_ID < 80000 && \LIBXML_VERSION < 20900) {
+            // Turn off the entity loader
+            $this->xmlLoaderValue = libxml_disable_entity_loader(true);
+        }
 
         // Suppress the errors because we don't really have to worry about formation before cleansing
         libxml_use_internal_errors(true);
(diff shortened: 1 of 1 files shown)


darylldoyle (dismissed):
This looks good to me

iamdharmesh:
Hi @jeffpaul ,

Changes in PR looks good to me. but changes are made in `lib/vendor/enshrined/svg-sanitize` folder, so, the best way is fix this issue at the package level and update the package here with `composer update`. 

Issue is already fixed in the composer package with this PR https://github.com/darylldoyle/svg-sanitizer/pull/45/files. So, better to fix this issue with `composer update`. but package requires a minimum PHP7.0 (https://github.com/darylldoyle/svg-sanitizer/blob/master/composer.json#L27) while this plugin has support for PHP 5.6 & greater.

So, can you please suggest on can we proceed with merge this PR or should go with updating the package?

darylldoyle:
Good catch @iamdharmesh! 

@jeffpaul I can check if the lib will run on 5.6 still, but I've got a feeling I dropped support to use new features. If it doesn't I wonder if it'd be worth upping the min PHP version for the plugin to PHP 7? Whilst [core still supports 5.6](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/), breaking away from the underlying library will create a huge amount of overhead on the project as we'd need to manually patch everything from upstream.

jeffpaul:
@darylldoyle @iamdharmesh bumping PHP minimum seems fine, though we'll want to clearly call that out in the changelog and in an upgrade notice in the readme.txt file.  Anyone running a lower PHP version should be blocked by WP core in attempting to update but we should still clearly call out the change so folks are aware.

iamdharmesh:
Thanks for feedback @jeffpaul @darylldoyle ,
I will close this PR and will create a new issue for updating the composer package and bumping PHP minimum version.

iamdharmesh:
Closing this PR as this fix will be addressed with #17

PR #19, opened 2022-01-07
Bug fixes relating to image dimensions

### LINKED ISSUE #15: Two bug fixes relating to image dimensions
I see you're possibly releasing a new version shortly, so I thought I'd highlight two bug fixes that really need to be in it. Given the plugin hasn't been updated in ages, I've been hacking them into versions I install, so don't want them to go back to being broken on upgrade.

a) https://wordpress.org/support/topic/1-9-9-doesnt-return-correct-width-and-height/

A missing string cast causes dimensions to be lost from some SVGs.

b) https://wordpress.org/support/topic/square-svgs-are-totally-broken-in-chrome/

A bug with srcsets breaks square SVGs entirely in Chrome.

mehulkaklotar:
### Requirements

- Remove double slashes in the file url when rendered
- A missing string cast causes dimensions to be lost from some SVGs.
- A bug with srcsets breaks square SVGs entirely in Chrome. - We will not fix it as it crosses with WP.

### Description of the Change

- For newer uploads we have fixed the double slash being added in svg file url
- Float value casting fix for svgs when fetching width and height.

### Alternate Designs

No alternate designs.

### Benefits

- SVG uploaded without viewbox will be rendered properly on front end.

### Possible Drawbacks

No drawback.

### Verification Process

- Upload SVG in Media and see that attachment meta has now proper file path which should be `2022/01/FILENAME.svg`
- Float cast of width and height when fetching svg sizes gives correct width and height when rendering.

### Checklist:

### Applicable Issues

#15

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -327,7 +327,7 @@ if ( ! class_exists( 'safe_svg' ) ) {
                 $svg_path               = get_attached_file( $attachment_id );
                 $upload_dir             = wp_upload_dir();
                 // get the path relative to /uploads/ - found no better way:
-                $relative_path = str_replace( $upload_dir['basedir'], '', $svg_path );
+                $relative_path = str_replace( trailingslashit( $upload_dir['basedir'] ), '', $svg_path );
                 $filename      = basename( $svg_path );
 
                 $dimensions = $this->svg_dimensions( $svg_path );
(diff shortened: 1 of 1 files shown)


darylldoyle (approved):
This looks good to me!

PR #20, opened 2022-01-07
Update enshrined/svg-sanitize and bump PHP minimum version to 7.0

### LINKED ISSUE #17: Update enshrined/svg-sanitize and bump PHP minimum version to 7.0
**Is your enhancement related to a problem? Please describe.**

The original issue can be found here #12 (PHP 8.0: libxml_disable_entity_loader function is deprecated)

**Describe the solution you'd like**

`enshrined/svg-sanitize` lib already has added support for PHP 8.0. So, need to update lib to add that support to the wp plugin.
`enshrined/svg-sanitize` requires minimum PHP 7.0. So, need to bump PHP minimum version to 7.0 with upgrade notice as per @jeffpaul 's comment [here](https://github.com/10up/safe-svg/pull/12#issuecomment-999610559)
> we'll want to clearly call that out in the changelog and in an upgrade notice in the readme.txt file. Anyone running a lower PHP version should be blocked by WP core in attempting to update but we should still clearly call out the change so folks are aware.

**Designs**
n/a

**Describe alternatives you've considered**

**Additional context**

Noting here for giving credit to https://github.com/amdd-tim for raising original PR and issue.(#12)

mehidi258:
Fix #17


iamdharmesh:
Hi @mehidi258, 

Thanks a lot for the PR. Changes in PR look good to me. Just 2 minor changes are required.
- Please add `Requires PHP` & `Requires at least` in Plugin Header (in safe-svg.php)
- Please add `Upgrade Notice` about PHP minimum version bumped to 7.0

Thanks.

iamdharmesh (dismissed):
Thanks for changes @mehidi258 .

Hi @jeffpaul, Just wanted to double confirm one thing. This PR has added `Requires PHP` and `Upgrage Notice` in readme.txt file. Considering the huge userbase of plugin, do you think we need to add some code level check for `PHP < 7.0`. something like prevent activate plugin if PHP version is lower than `7.0` or show warning etc...? If not then we can proceed with merging this PR.

Thanks

jeffpaul:
@iamdharmesh if there's no check in core on php version then yes we should likely add what you've suggested

mehidi258:
> @iamdharmesh if there's no check in core on php version then yes we should likely add what you've suggested

@jeffpaul @iamdharmesh  There is already some checks

image

iamdharmesh (approved):
Thanks for checking @mehidi258. @jeffpaul all good to go.

PR #35, opened 2022-02-22
Update/composer install lib

darylldoyle:
### Description of the Change

* This moves the composer installed library to the root and makes sure that we're not including the entire library within the repo.
* It also updates the build steps to run `composer install` rather than `npm run build`.

### Checklist:

### Changelog Entry
* Added - New build step
* Changed - Library location


jeffpaul:
@dkotter this is probably worth a review to get into the 1.9.10 release to ensure things work smoothly as our first release under 10up / using our GitHub Actions.

.github/workflows/wordpress-plugin-deploy.yml
```diff
@@ -13,8 +13,7 @@ jobs:
       uses: actions/checkout@v2
     - name: Build
       run: |
-        npm install
-        npm run build
+        composer install
```
dkotter, line 16: It looks like we have a similar build command in the [wordpress-plugin-asset-update.yml](https://github.com/10up/safe-svg/blob/develop/.github/workflows/wordpress-plugin-asset-update.yml#L14) file. Should that be updated as well?

safe-svg.php
```diff
@@ -14,7 +14,7 @@
 
 defined( 'ABSPATH' ) or die( 'Really?' );
 
-require 'lib/vendor/autoload.php';
+require 'vendor/autoload.php';
```
dkotter: Since we aren't committing this file to the repo anymore, I think we need to add a conditional around this so those that are directly checking this repo out don't end up with a fatal error if they haven't run `composer install`.

Something like:
```suggestion
// Try and include our autoloader.
if ( is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
    require __DIR__ . '/vendor/autoload.php';
} else {
    add_action( 'admin_notices', function() { ?>
        <div class="notice notice-error">
            <?php /* translators: %1$s: composer command */ ?>
            <p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of Safe SVG. Please run %1$s in order for things to work properly.', 'safe-svg' ), '<code>composer install</code>' ) ); ?></p>
        </div>
<?php
    } );
    return;
}
```

dkotter (changes requested):

dkotter:
Should probably also add `vendor/*` into our `.gitignore` file

.github/workflows/wordpress-plugin-deploy.yml
darylldoyle: Thanks! I missed that other build command. That's been updated now!

safe-svg.php
darylldoyle: Good call, I've added this in 🙂

darylldoyle:
@dkotter, thanks for taking a look. I've made those updates to the PR!

dkotter (approved):

PR #46, opened 2022-03-16
Ensure we have a proper file path and file name before we use them

### LINKED ISSUE #45: Undefined $file['tmp_name']
### Describe the bug

When I submit a form from another extension, there is an empty file entry.
I have an error generated by safe-svg because there is an undefined variable.

In function `check_for_svg` :

`$wp_filetype = wp_check_filetype_and_ext( $file['tmp_name'], $file['name'] );`

If I add this code before, it's working !

```
if ( ! isset( $file['tmp_name'] ) || ! isset( $file['name'] ) ) {
    return $file;
}
```

### Steps to Reproduce

Maybe create a form with empty file input.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

As reported in #45, if for some reason a file `tmp_name` (the temporary location of the uploaded file) or file `name` (the actual file name) doesn't exist in the file array, a PHP error can be generated because we are accessing an undefined variable. This shouldn't be a common scenario (because if the `tmp_name` doesn't exist, that means the file wasn't uploaded properly) but is something we should try catching.

This PR fixes this be ensuring both the file `tmp_name` and file `name` exist before we use those.

Closes #45

### Possible Drawbacks

None

### Verification Process

I was not able to reproduce this myself but the scenario is you somehow upload a file but the upload fails and yet things still keep processing.

### Checklist:

### Changelog Entry

> Fixed - Ensure the `tmp_name` and `name` properties exist before we use them

### Credits

Props @aksld

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -133,7 +133,13 @@ if ( ! class_exists( 'safe_svg' ) ) {
          */
         public function check_for_svg( $file ) {
 
-            $wp_filetype = wp_check_filetype_and_ext( $file['tmp_name'], $file['name'] );
+            // Ensure we have a proper file path before processing
+            if ( ! isset( $file['tmp_name'] ) ) {
+                return $file;
+            }
+
+            $file_name   = isset( $file['name'] ) ? $file['name'] : '';
+            $wp_filetype = wp_check_filetype_and_ext( $file['tmp_name'], $file_name );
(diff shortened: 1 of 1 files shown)


aksld:
Perfect !

cadic (approved):
LGTM!

PR #52, opened 2022-04-08
Check if Sanitizer class exists before bailing

### LINKED ISSUE #40: Future plans for composer package
Hey, great to see 10up taking on maintenance here! I was wondering what if any plans you might have for the future in terms of this plugin being on packagist.org. Will you be updating the name and resubmitting?

Also I noticed that 1.9.10 seems like it shouldn't really be a patch release if there are breaking changes with regards to WP and PHP version support. Not suggesting removing that tag of course but wanted to note it could cause problems for people using a loose enough version constraint.

roborourke:
### Description of the Change

Installs via packagist will have the required `enshrined/svg-sanitize` dependency in the main project's `vendor` directory. This change will allow such installs to continue working, providing that the application loads the project level autoloader before plugins. Such a pattern is used in frameworks like Roots Bedrock and Altis.

Related to #40 

### Alternate Designs

N/A

### Possible Drawbacks

There might be some unusual situations in which dependencies are installed and loaded where the version of `enshrined/svg-sanitize` is not what is required by this plugin, though it would require quite a convoluted set up for that to happen so it seems low risk to me.

### Verification Process

- Install plugin via packagist, e.g. `composer require darryldoyle/safe-svg`
- Apply this patch to the plugin 
- Activate plugin
- Check to see if admin notice displayed
- Check to see if uploading and displaying an SVG containing `<script>alert('XSS');</script>` fires the alert

### Checklist:

### Changelog Entry

> Fixed: Support for installing via packagist.org

### Credits

Props @roborourke

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -19,7 +19,7 @@ defined( 'ABSPATH' ) or die( 'Really?' );
 // Try and include our autoloader.
 if ( is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
 	require __DIR__ . '/vendor/autoload.php';
-} else {
+} elseif ( ! class_exists( 'enshrined\\svgSanitize\\Sanitizer' ) ) {
 	add_action(
 		'admin_notices',
 		function() {


peterwilsoncc (approved):
Code looks good to me, verified with manual testing.

**Testing notes**

I couldn't figure out how to require the PR commit of this via composer so needed to test in a round about fasion

* checked out this branch in my plugins folder without running composer install
* Activated plugin
* Developer version notice displayed
* Added `enshrined/svg-sanitize` as a dependency of my local site (ie, not the plugin)
* required `vendor/autoload.php` in my `wp-config.php` file
* Added `define( 'ALLOW_UNFILTERED_UPLOADS', false );` to `wp-config.php`
* Reloaded site's admin
* Notice did not display
* confirmed I could upload the SVG ✅

PR #63, opened 2022-05-25
Remove CSS fix that is no longer needed

### LINKED ISSUE #62: Extra space in Featured Image
### Describe the bug

When adding an SVG as a Featured Image, there is extra padding that appears above the featured image.

This seems to be related to some CSS that is added to this plugin.

### Steps to Reproduce

1. Go to any page.
2. Add an SVG as a Featured Image
3. Note the extra space.

Without the plugin enabled:
Screen Shot 2022-05-17 at 8 58 09 AM

With the plugin enabled:
Screen Shot 2022-05-17 at 8 57 39 AM

The problematic code in question:
```
/**
 * Fix for Gutenberg not showing featured images correctly.
 */
img.components-responsive-wrapper__content[src$=".svg"] {
    position: relative;
}
```

### Environment information

- WordPress 5.9.3
- Chrome 101.0.4951.64

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

As described in #62, there's some unnecessary spacing being added to the display of featured images in the admin, when an svg is used. This is because of a line of CSS we were adding. This PR removes that CSS.
 
I wanted to trace the history of that to ensure we could safely remove that. It seems that was added in the [v1.8.1 release](https://github.com/10up/safe-svg/commit/770998d0ab7531b131207dd0e39fe1bf66d667f0#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6R67) (Nov 2018) and was mentioned to be a fix for the featured image display in Gutenberg. Looking further, the issue was reported on the [.org forums](https://wordpress.org/support/topic/svgs-in-gutenberg-editor/) just a week or so before that.

After removing that CSS, I checked a WP 5.0 environment and could not replicate the original reported issue. I'm assuming this was fixed in Gutenberg prior to the WP 5.0 release (which was only a month or so after this fix was added to Safe SVG). So as far as I can tell, we are safe to remove this. 

Closes #62

### Note

While testing this, I was able to reproduce another scenario where extra spacing was being added to the featured image display in the admin. Diving into the code that Gutenberg uses to render this, I found there's a [Responsive Wrapper component](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/responsive-wrapper/index.js) that adds a span tag above the featured image tag and adds an inline padding-bottom to that. This padding is based on the height and width of the featured image.

In the version of Gutenberg that shipped with WP 6.0, we default to the `post-thumbnail` image size, if that exists. In the most recent version of the Gutenberg plugin (and what I assume will ship in WP 6.1) we now default to the [`large` image size](https://github.com/WordPress/gutenberg/pull/40126/files#diff-9b8c234ca3c82c5dcc1841899d5825553a4fd559b4c7d03c96a827b0ecbff314R58).

I'm running the Twenty Twenty One theme locally and it has a `post-thumbail` size set to [1568x9999](https://github.com/WordPress/twentytwentyone/blob/trunk/functions.php#L72). When using a size of 9999, WordPress is supposed to use the other dimension first and then scale the second appropriately. When an SVG is uploaded, Safe SVG tries to store the proper image sizes in meta (without actually generating re-sized images) and so we end up storing the size 1568x9999 (since SVGs can scale infinitely).

This ends up being the image dimensions that are used for the padding calculation mentioned above and results in something like 600%+ bottom padding being added. If I switch to the latest version of the Gutenberg plugin (and thus it now uses the `large` image size), the bottom padding is calculated correctly.

I think ultimately this is another bug we probably should look to fix here, that when a site has an image size registered to be 9999, instead of just using that as-is, we should try and figure out the scaled dimensions to use instead. Probably not a big deal to fix but mentioning here as other's may run into the same issue I saw where extra spacing is still being added even after this fix.

### Alternate Designs

None

### Possible Drawbacks

Potential to introduce a regression that this line of CSS was added to solve in the first place. I couldn't replicate any problems myself though

### Verification Process

1. Add an SVG as a featured image to a post
2. Verify there isn't extra spacing being added

### Checklist:

### Changelog Entry

> Removed - unneeded admin CSS fix

### Credits

Props @AdamWills, @dkotter

### DIFF
--- a/assets/safe-svg.css
+++ b/assets/safe-svg.css
@@ -4,10 +4,3 @@
 #postimagediv .inside .svg img {
     width: 100%;
 }
-
-/**
- * Fix for Gutenberg not showing featured images correctly.
- */
-img.components-responsive-wrapper__content[src$=".svg"] {
-    position: relative;
-}


peterwilsoncc (commented):
I am not sure if this is a bug with the plugin or the block editor but I am seeing some big margins when opening a saved post with an SVG featured image.

I am testing with 2022 active.

When the image is first placed the margins do not exist.

svg-featured

dkotter:
> I am not sure if this is a bug with the plugin or the block editor but I am seeing some big margins when opening a saved post with an SVG featured image.

I'm not seeing this myself, tested with both the 2021 and 2022 themes and with the Gutenberg plugin active and not active. As mentioned in the PR description, I can see an issue when not using the latest Gutenberg plugin and having a theme active that has a `post-thumbnail` size of some-widthX9999 (which both 2021 and 2022 do). But the extra spacing is there both on initial load and after setting a featured image, which it sounds like you only see it on initial load.

I'm not seeing it in your gif but if you inspect the actual featured image tag, you should see a `<span>` tag right above that has a `padding-bottom` applied. In the tests I ran with a certain SVG and without Gutenberg active (so it defaults to using the `post-thumbnail` size) I'm seeing a padding of 637%. With Gutenberg active (so it now uses the `large` image size) it gives a padding of 100%, which addresses the issue. I'm assuming in your case it's applying different paddings on initial load compared to when setting a featured image, which I'm not sure myself if that's an editor issue or an issue here. I wouldn't be surprised if it has something to do with how we store the various image sizes for svgs but I would also expect that to impact both scenarios the same.

peterwilsoncc:
> I'm assuming in your case it's applying different paddings on initial load compared to when setting a featured image,

Yes, that's what is happening. On page load it shows `<span style="padding-bottom: 100%;"></span>`, on placing the image it shows `<span style="padding-bottom: 150px;"></span>`. This is the image I am testing with.

sample-placeholder

I was seeing some strange behaviour with `9999` images so experimented with changing the resizing loop in `skip_svg_regeneration()` and it seemed to work better. I wonder if we're seeing different issues.

```php
foreach ( wp_get_registered_image_subsizes() as $name => $settings ) {
    $new_dimensions = $dimensions;
    if (
        $settings['width'] > $dimensions['width'] ||
        $settings['height'] > $dimensions['height']
    ) {
        $multiplier_w = $settings['width'] / $dimensions['width'];
        $multiplier_h = $settings['height'] / $dimensions['height'];

        $multipier = $multiplier_w > $multiplier_h ? $multiplier_w : $multiplier_h;
        $new_dimensions['width'] = floor( $dimensions['width'] * $multipier );
        $new_dimensions['height'] = floor( $dimensions['height'] * $multipier );
    }

    if ( $settings['width'] >= 9999 || $settings['height'] >= 9999 ) {
        $settings['crop'] = false;
    }


    $resized = image_resize_dimensions( $new_dimensions['width'], $new_dimensions['height'], $settings['width'], $settings['height'], $settings['crop'] );

    $sizes[ $name ]['width'] = $resized[6];
    $sizes[ $name ]['height'] = $resized[7];
    $sizes[ $name ]['file']      = $filename;
    $sizes[ $name ]['mime-type'] = $mime;
    $sizes[ $name ]['crop'] = intval( $settings['crop'] );

}
```

AdamWills:
This indeed fixes the issue on our end.

Before:
Screen Shot 2022-05-27 at 9 21 26 AM

After:
Screen Shot 2022-05-27 at 9 24 20 AM

Both of these screen shots are from initial loads of a page, so I'm not seeing what @peterwilsoncc is seeing.

dkotter:
@AdamWills Thanks for testing!

@peterwilsoncc I am able to reproduce the same thing with your example SVG, though here's what I'm seeing:
1. When I first set that SVG as the featured image, it has extra spacing above and below
2. If I refresh the edit screen, I see the same spacing
3. If I then use the `Replace Image` button, change to a non-SVG image and then back to the same SVG I had originally, the extra spacing isn't there
4. If I instead use the `Remove featured image` button to completely remove it and then add the same SVG back as the featured image, I still have the extra spacing

So not sure if it's something with that particular SVG or if there's an issue with the sizes we set in `skip_svg_regeneration` or if this is a block editor issue (or some combination of those). But I don't think this is related with the change in this PR so unless you feel differently, I think we can merge this fix in and open a separate issue (if we think it's worth investigating more) to continue looking into this

AdamWills:
@dkotter - it does seem like the issue above is a separate issue.

peterwilsoncc (approved):
> But I don't think this is related with the change in this PR so unless you feel differently, I think we can merge this fix in and open a separate issue (if we think it's worth investigating more) to continue looking into this

Cool, that works for me.

PR #76, opened 2022-08-22
add settings for user role selection

### LINKED ISSUE #69: Choose Who Can Upload

Port in from [Pro version](https://github.com/10up/wpsvg)

### Designs

_No response_

_No response_

### Code of Conduct

dhanendran:
### Description of the Change
Added a settings section to select user roles who can upload SVG images. Based on the role section, a new capability `safe_svg_upload_svg` will be added to those roles. By default, super admins on a network site will have all the capabilities, so they can upload SVG files.

Closes #69 

### How to test the Change
1. Go to `/wp-admin/uploads.php` page
2. Try to upload an SVG file and it should not be uploaded and get an error message
3. Enable the plugin and go to `/wp-admin/options-media.php` page
4. All the available roles on the site should be listed with the checkbox
5. Admin should be able to select the user roles who can upload the SVG files
6. Select `Administrator` role and submit the form
7. Go to `/wp-admin/uploads.php` page
8. Try to upload an SVG file and it should be uploaded successfully

### Changelog Entry
Feature - Add a new option to select which user roles can upload SVG files.

### Credits
Props @dhanendran, @csloisel, @faisal-alvi 

### Checklist:


README.md
```diff
@@ -8,15 +8,14 @@
 
 Safe SVG is the best way to Allow SVG Uploads in WordPress!
 
-It gives you the ability to allow SVG uploads whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views.
+It gives you the ability to allow SVG uploads whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views. It gives you the ability to select which user role can upload an SVG file.
```
dkotter:
```suggestion
It gives you the ability to allow SVG uploads (and limit that by user role) whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views.
```

readme.txt
```diff
@@ -14,15 +14,14 @@ Enable SVG uploads and sanitize them to stop XML/SVG vulnerabilities in your Wor
 
 Safe SVG is the best way to Allow SVG Uploads in WordPress!
 
-It gives you the ability to allow SVG uploads whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views.
+It gives you the ability to allow SVG uploads whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views.  It gives you the ability to select which user role can upload an SVG file.
```
dkotter:
```suggestion
It gives you the ability to allow SVG uploads (and limit that by user role) whilst making sure that they're sanitized to stop SVG/XML vulnerabilities affecting your site.  It also gives you the ability to preview your uploaded SVGs in the media library in all views.
```

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,101 @@
+<?php
+/**
+ * Safe SVG Plugin settings.
```
dkotter:
```suggestion
 * Safe SVG plugin settings.
```

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,101 @@
...
+	 * @param array $args The settings array, defining title, id, callback.
+	 */
+	public function safe_svg_settings_callback( $args ) {
+		?>
+		<p id="<?php echo esc_attr( $args['id'] ); ?>"><?php esc_html_e( 'Select user roles who can upload SVG files.', 'safe-svg' ); ?></p>
+		<?php
+	}
+
+	/**
+	 * User role field callbakc function.
```
dkotter:
```suggestion
	 * User role field callback function.
```

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * Set up the class
+	 */
+	public function __construct() {
+		add_action( 'admin_init', [ $this, 'settings_init' ] );
+		add_filter( 'pre_update_option_safe_svg_upload_roles', [ $this, 'update_capability' ], 10, 2 );
+	}
+
+	/**
+	 * custom option and settings
```
dkotter:
```suggestion
	 * Custom option and settings
```

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,103 @@
...
+			<?php esc_html_e( 'Select user roles who can upload SVG files.', 'safe-svg' ); ?>
+		</p>
+		<?php
+	}
+
+	/**
+	 * User role field callback function.
+	 */
+	public function safe_svg_roles_cb() {
+		$user_roles   = get_editable_roles();
```
dkotter: This will show all roles, even those that can't do anything with media currently. Feels like we should trim this list down to only roles that currently have media upload capabilities to begin with, otherwise I could see confusion when someone sees `Subscriber` on this list, even though they can't upload any types of media

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,103 @@
...
+		$upload_roles = (array) get_option( 'safe_svg_upload_roles', [] );
+
+		foreach ( $user_roles as $role => $info ) :
+			?>
+			<div>
+				<label>
+					<input type="checkbox" name="safe_svg_upload_roles[]" value="<?php echo esc_attr( $role ); ?>" <?php checked( in_array( $role, $upload_roles, true ), true ); ?> /> <?php echo esc_html( $info['name'] ); ?>
```
dkotter, line 123: The way this is currently set up, no roles will be selected by default, meaning after this update, site admins will need to go in and select the proper roles before things work again (which I think is going to lead to support requests and complaints).

I would suggest that Admins and probably Editors be selected by default. This makes things a little tricky, as we would need to distinguish between situations where this setting hasn't been saved yet and where someone has purposefully deselected all roles and saved that value.

I _think_ the easiest approach is something like:

1. Modify our `register_setting` call to add a `sanitize_callback` function
2. Within that callback, check that we have an array of data and if so, we should probably do something like `array_map( 'sanitize_text_field', $roles )`. If the data we have isn't in an array (so if no roles are selected, I think the value passed will be `null`) we instead save something like `none`
3. We'll need to modify the `update_capability` method to account for that `none` string, basically a check right at the beginning of that method that returns early if `$new_roles === 'none'`
4. And then instead of just doing `current_user_can( 'safe_svg_upload_svg' )`, we'll need a custom method that checks what `get_option( 'safe_svg_upload_roles' )` returns. If it's empty, we can assume that setting hasn't been initialized yet and we can then check if the current user is an admin or editor to determine if they should have access. If it returns an array value or the value `none`, we can do the normal `current_user_can` check

dkotter (changes requested):

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Settings section callback function.
+	 *
+	 * @param array $args The settings array, defining title, id, callback.
+	 */
+	public function safe_svg_settings_callback( $args ) {
+		?>
+		<p id="<?php echo esc_attr( $args['id'] ); ?>">
+			<?php esc_html_e( 'Select user roles who can upload SVG files.', 'safe-svg' ); ?>
```
dkotter:
```suggestion
			<?php esc_html_e( 'Select which user roles can upload SVG files.', 'safe-svg' ); ?>
```

csloisel:
@dkotter I've added a bunch of stuff based on your feedback:
https://github.com/10up/safe-svg/pull/76/files/5c8fa7068164bb60593987a5e8052352c57d0391..37b28c739807752fc399579c6ca7fc609163bec5

- Updated branch from develop
- Filtered role options down to only ones that can edit media to being with
- Adding helper method to get upload cap roles
- Add sanitize function for roles
- Checking roles are role objects before calling methods
- Use upload roles as default if nothing is stored in settings
- Replacing current_user_can with custom method to check for empty settings

includes/safe-svg-settings.php
csloisel: I've updated this in the latest push, it will be filtered to roles that have the upload_files permission. I've also added a filter for these roles.

includes/safe-svg-settings.php
csloisel: I have added the functionality you specced out here, however, I have made all upload roles selected by default for full backwards compatibility.

jeffpaul:
@dkotter looking for an update code review from you here, thanks!

jeffpaul:
@faisal-alvi if you're able to review during your OSS week that'll help move this along, thanks!

includes/safe-svg-settings.php
```diff
@@ -0,0 +1,154 @@
...
+	public function get_upload_capable_roles() {
+		$all_roles    = get_editable_roles();
+		$upload_roles = array_filter(
+			$all_roles,
+			function( $_role ) {
+				return $_role['capabilities']['upload_files'] ?? false;
+			}
+		);
+
+		return apply_filters( 'safe_svg_upload_roles', $upload_roles, $all_roles, $this );
```
dkotter, line 92: Would be great to have a docblock around this filter

dkotter:
> @faisal-alvi if you're able to review during your OSS week that'll help move this along, thanks!

I've reviewed the code here a few times and things are looking good from my perspective. Could use help in actually testing this out, want to ensure this works as expected and doesn't introduce any regressions. The main thing I'm concerned about is ensuring that anyone that can currently upload an SVG should still be able to upload an SVG after this update, at least until this new setting has been modified from the default.

Edit: looks like also a merge conflict that we can clean up here

faisal-alvi (dismissed):
@dhanendran & @csloisel thank you for the great work here!

Note: this is a report of testing only, as the Code Review is already been done by @dkotter.

* ✅ A settings section to select user roles who can upload SVG images is displayed correctly.

image

* ✅ I have tested and confirmed the functionality is working fine.
* ✅ Only the user roles allowed in settings can upload the SVG.
* ✅ If a user with a specific role is not allowed to upload, the following error is displayed and the SVG image does not upload.

image

jeffpaul:
Resolved the merge commit, but seeing PHPUnit failure that we'll want to resolve before merging

jeffpaul:
@10up/open-source-practice would be good to get a final code review pass on this before merging in... thanks!

dkotter (approved):

PR #79, opened 2022-10-09
Feature/svgo support

### LINKED ISSUE #68: SVGO Optimisation

Port in from [Pro version](https://github.com/10up/wpsvg)

### Designs

_No response_

_No response_

### Code of Conduct

gsarig:
### Description of the Change

Implements support for SVG optimization using the popular [SVGO tool](https://github.com/svg/svgo). For the scripts' structure and build, [10up-toolkit](https://github.com/10up/10up-toolkit) has been utilized. The Optimizer is active by default and will try to optimize any SVG file uploaded via the Media Library, using the default SVGO settings.
 
A filter `safe_svg_svgo_param` has been added so that developers can override these settings and pass their own. 
Also, using the same filter like this: `add_filter( 'safe_svg_svgo_params', '__return_false' );` will allow them to completely disable the Optimizer. 

~~At the moment, the optimizer doesn't work if you upload an image directly (using "Upload" instead of the "Media Library" button), as it bypasses the media uploader.~~ Other than that, as long as the file is uploaded using the Media Library, it will get optimized. Also, it should work with both single and multiple images. With the first chance, I will resume my work to make it work with direct uploads as well. 

Closes https://github.com/10up/safe-svg/issues/68 

### How to test the Change

1. Run `npm install` and `npm run build`.
2. Upload one or more unoptimized SVG files using the Media Library. You can find unoptimized SVG files for testing purposes [here](https://unoptimized--svg-icon-stress-test.netlify.app/).  
3. Confirm that the uploaded file(s) have been optimized. This can be checked by either comparing their file size or by adding the files in a text editor and comparing their contents (optimized files should have considerably less content). 

Known issue: Given that SVGO is a JS tool, using it with WordPress had its challenges. To make it work, a hook to `wp.Uploader` has been made, to identify when a file gets uploaded, optimize it, and then make an AJAX call to update the file with PHP. ~~Therefore, at the moment it doesn't work if you upload an image directly, bypassing the media uploader. More specifically, if you hit the "Upload" button from the image below:~~
image
~~Using the "Media Library" option, on the other hand, should work with either single or multiple SVG. It should also work with drag and drop and even if you drop mixed image types (for example SVGs together with JPEGs, PNGs etc) at the same time.~~

Update: Support for optimizing on direct uploads has been added [on this commit](https://github.com/10up/safe-svg/pull/79/commits/bcc12c2262d8e875877f14f6e378f67ff8c2da15), and images uploaded using the direct upload button will get optimized when the post is saved.  

### Changelog Entry

> Added - SVG optimization during upload via SVGO

### Credits

Props @gsarig

### Checklist:


jeffpaul:
@Sidsector9 @darylldoyle looking for review from you both on this, thanks!

Sidsector9 (dismissed):
Thank you very much for working on this PR.

- Tested by uploading a 1KB SVG using the Media Uploader. The size got reduced to 411 bytes.
- Tested the filter `add_filter( 'safe_svg_svgo_params', '__return_false' );` to disable optimisation, works as ecpected.

This is the image I used:

sample

Sidsector9:
@jeffpaul I am wondering if we can extend this to a separate feature to support a batch optimiser, which will optimise already uploaded SVG images, what do you think?

Sidsector9 (changes requested):
Thank you very much for working on this PR.

✅  Tested by uploading a 1KB SVG using the Media Uploader. The size got reduced to 411 bytes.
✅  Tested the filter `add_filter( 'safe_svg_svgo_params', '__return_false' );` to disable optimisation, works as excpected.
❌ Direct upload using core/image block did not optimise the SVG

This is the image I used:

sample

Please watch the video below:

https://user-images.githubusercontent.com/17757960/196622537-e7f27a56-e150-4cff-8a8f-df1b2332ffb6.mov

gsarig:
> Thank you very much for working on this PR.
> 
> ✅ Tested by uploading a 1KB SVG using the Media Uploader. The size got reduced to 411 bytes. ✅ Tested the filter `add_filter( 'safe_svg_svgo_params', '__return_false' );` to disable optimisation, works as excpected. ❌ Direct upload using core/image block did not optimise the SVG
> 
> This is the image I used:
> 
> sample
> 
> Please watch the video below:
> 
>  SVGO.mov

Hi @Sidsector9 Thanks for testing this. This is something that I should have mentioned in the PR (I've updated the description): the image will get optimized on post save. So, if you click to update the post, the image should get optimized.

Sidsector9 (dismissed):
Thanks for the additional details @gsarig. LGTM, approving 👍

gsarig:
> @jeffpaul I am wondering if we can extend this to a separate feature to support a batch optimiser, which will optimise already uploaded SVG images, what do you think?

@Sidsector9 @jeffpaul My thought here was to allow it to optimize existing SVGs when someone triggers a batch regenerate of images, with WP-CLI or a third-party plugin. That way we wouldn't have to build a settings page for that, or a specific mechanism from scratch and the implementation should be quicker. 
The challenge with the optimizer is that being a JS tool, it cannot utilize the standard PHP hooks so it would take some testing/investigation to see how this could best work.

Sidsector9:
@gsarig yeah a feature like that would be a good addition to the plugin's capabilities. We can open a separate issue to investigate how best to approach this.

jeffpaul:
@vikrampm1 can you please open an issue as noted by Sid and Giorgos above?

assets/js/admin/admin.js
```diff
@@ -0,0 +1,117 @@
...
+        if (isSavingPost()) {
+            let changes = getPostEdits();
+            for (const changedBlock of changes.blocks) {
+
+                // Run only if this is a new core image block.
+                if ('core/image' === changedBlock.name && !changedBlock?.originalContent) {
+                    const svgUrl = changedBlock?.attributes?.url;
+
+                    // Run only if the image is an SVG that has not been optimized yet.
+                    if (!svgUrl || !svgUrl.endsWith('.svg')) {
```
darylldoyle: **thought (non-blocking):** This will only pick up files with the `.svg` extension. What about `.svgz` or SVGs that are uploaded without the SVG extension? 

Do we have access to the mime type at this point to check that instead?

assets/js/admin/admin.js
```diff
@@ -0,0 +1,117 @@
...
+     */
+    const {isSavingPost, getPostEdits} = select('core/editor');
+
+    subscribe(() => {
+        if (isSavingPost()) {
+            let changes = getPostEdits();
+            for (const changedBlock of changes.blocks) {
+
+                // Run only if this is a new core image block.
+                if ('core/image' === changedBlock.name && !changedBlock?.originalContent) {
```
darylldoyle: **thought (non-blocking):** This only works for `core/image` blocks. What about `core/gallery`, `core/media-text` etc.

assets/js/admin/admin.js
```diff
@@ -0,0 +1,117 @@
...
+    );
+
+    /**
+     * Optimize the SVGs inserted in the editor.
+     * This takes care of SVGs uploaded via direct upload, without using the Media Library.
+     */
+    const {isSavingPost, getPostEdits} = select('core/editor');
+
+    subscribe(() => {
+        if (isSavingPost()) {
```
darylldoyle: **thought (non-blocking):** The handling of this on save concerns me a little. We're making 2 additional requests for each image on the page. That could add up quickly and really bog down the save process of the block editor. 

Have we looked other options, such as something like https://github.com/spatie/image-optimizer which _may_ allow us to do the optimisation from PHP at upload time.

jeffpaul:
@gsarig some input from Daryll above, let me know if you hope to be able to react to those or if I should pull someone additional to help finish this off?

gsarig:
> @gsarig some input from Daryll above, let me know if you hope to be able to react to those or if I should pull someone additional to help finish this off?

Hi @jeffpaul 
I saw Daryll's input - I will finish this, probably this weekend.

assets/js/admin/admin.js
gsarig: @darylldoyle We don't (that's the reason I made the check based on the filename). Here's an example of what the changedBlock returns:

image

One approach that I can think of is to remove that particular check and perform it inside the `fetch`, where the file's `content-type` becomes available. In that case, though, the `fetch` would have to run on every file, while now we only run it on .`svg` files. Perhaps we could mitigate that by narrowing it down to specific extensions (e.g. `svg*` for anything that ends in svg-something) and files with no extension at all. That way, we wouldn't bother running it on jpgs, pngs etc.

What do you think?

assets/js/admin/admin.js
gsarig: @darylldoyle I've pushed a new commit, which adds support for `core/media-test` and for nested blocks containing images (which covers the `core/gallery` scenario). 
Can you please take a look? 
Thanks!

assets/js/admin/admin.js
gsarig: @darylldoyle I agree with your arguments, and this is something that concerned me as well before I ended up with the specific approach. When testing the `image-optimizer`, though, I came across the following issues, which made me avoid it:
1. It uses SGVO 1, which is the older version of SVGO, and based on that [comment](https://github.com/spatie/image-optimizer/issues/147), it seems to have a vulnerable dependency.
2. When I tried to test it, I got a fatal error, as one of the tool's dependencies seem to require PHP 8.1.0. (The error was `Your Composer dependencies require a PHP version ">= 8.1.0".`. 

I do agree, that a PHP-based approach would be ideal, and much simpler to implement. I didn't find a reliable tool to use instead, though.

jeffpaul:
@darylldoyle @Sidsector9 looking to either for a review on the updates here

jeffpaul:
@10up/open-source-practice could use some help here on code review (as well as resolving merge conflicts)... thanks!

Sidsector9:
@gsarig after resolving conflicts I saw console errors on the media page in the `admin.js`. It said `select` not found.
As there has been a shift from wp-scripts (develop branch) to 10up-toolkit, would you mind resolving the conflicts? Once done I can review and merge this PR.

Thanks in advance!

gsarig:
> @gsarig after resolving conflicts I saw console errors on the media page in the `admin.js`. It said `select` not found. As there has been a shift from wp-scripts (develop branch) to 10up-toolkit, would you mind resolving the conflicts? Once done I can review and merge this PR.
> 
> Thanks in advance!

Hi @Sidsector9 I've updated the scripts and pushed the changes. There should be no conflicts anymore and you can review the PR. 

Let me know if you need anything else. 
Thanks!

jeffpaul:
@faisal-alvi if you're able to help with code review during your OSS week, then that will help move this along... thanks!

faisal-alvi (changes requested):
@gsarig thank you for your great work! I have tested and found the following cases.

The majority of them appear to have already been reported and, most likely, fixed, however, they keep regenerating for me. I have verified twice that the project is properly built (`npm install` and `npm run build`). Please check, and let me know if there is anything I missed.

* ✅ Works when being uploaded from the `Media > Library` section.
* ❌ Does not work when uploaded from the `Media > Add New > (browser uploader; input field)`
* ❌ Does not work when uploaded from the `Media > Add New > (multi-file uploader; drag and drop)`
* ❌ Does not work with the core image block. (the post was saved before checking the size)
* ❌ Does not work with the core gallery block. (the post was saved before checking the size)
* ❌ Does not work with the Safe SVG block. (the post was saved before checking the size)

https://user-images.githubusercontent.com/25176325/218765227-d93dcdcb-365a-46ad-8f57-5767c9385554.mp4

---

* ❌ Noticed the same console error that was reported earlier.

image

gsarig:
Thank you for the great feedback @faisal-alvi 

I pushed some updates to address the following issues:
1. Fix the console error `Uncaught TypeError: Cannot read properties of undefined (reading 'select')`
2. When uploading from the "Upload New Media" (using the "Add New" link on the Media library), the image didn't get optimized, as it initially only worked on the post editor screen. I pushed an update to fix it. It was a bit of a challenge because we have to rely on JS to perform the optimize, and we have different modes (Grid/List view and browser/multi-file uploader) but on my local tests it seems to work now. Keep in mind that the only scenario that this will not work is if you toggle the uploader and upload an image without a page reload first. The reason is that because the two uploaders have different logic and markup and need different approach in running the optimizer, in order for the script to work, it would need a page reload first to pick up the new markup. This is something that can be fixed, of course, but I will need to find some more time to do so. 
4. I also identified another issue: when I tried to upload an image on a page that wasn't yet published or saved, the optimizer wouldn't run. That's because initially it was set to run only on the `post.php` and not on the `post-new.php`, which is the edit screen of a new post. The latest updates should fix that as well. 

Can you please check again now and let me know what you think? 

Thank you

assets/js/admin/admin.js
```diff
@@ -0,0 +1,255 @@
...
+		}
+
+		// Prepare the parameters for the AJAX Call.
+		return {
+			action: 'safe_svg_optimize',
+			svg_url: svgUrl,
+			optimized_svg: optimizedString,
+			svg_nonce: safeSvgParams.nonce, // eslint-disable-line no-undef
+		};
+	};
```
faisal-alvi: Seems like a vertical alignment is required.

assets/js/admin/admin.js
```diff
@@ -0,0 +1,255 @@
...
+									if (ajaxResponse?.status !== 200) {
+										return;
+									}
+                                    // Remove the cookie.
+									document.cookie = safeSvgCookie + '=;expires=Thu, 01 Jan 1970 00:00:00 UTC;' + safeSvgCookieAttr;
+								});
+						});
+				}
+			}, 500);
+
```
faisal-alvi: Vertical alignments are required. Also, the comments also have spaces instead of tabs.

https://github.com/10up/safe-svg/blob/dc6dbdb7a47b930c54e70bc34f74140b27ac06be/assets/js/admin/admin.js#L41-L82

assets/js/admin/admin.js
```diff
@@ -0,0 +1,255 @@
...
+                                        wp.media.frame.content.get().options.selection.reset();
+                                    }
+                                } else {
+                                    wp.media.frame.library.props.set({ ignore: +new Date() });
+                                }
+                            });
+                    });
+            },
+        });
+    }
```
faisal-alvi: Vertical alignments are required. Also, the comments also have spaces instead of tabs.

https://github.com/10up/safe-svg/blob/dc6dbdb7a47b930c54e70bc34f74140b27ac06be/assets/js/admin/admin.js#L84-L172

webpack.config.js
```diff
@@ -10,5 +10,6 @@ module.exports = {
     entry: {
         'safe-svg-block': [ './includes/blocks/safe-svg/index.js' ],
         'safe-svg-block-frontend': [ './includes/blocks/safe-svg/frontend.js' ],
+        'safe-svg-admin': [ './assets/js/admin/admin.js' ],
```
faisal-alvi, line 13: Should be this file added to the `.distignore`?

faisal-alvi (dismissed):
@gsarig thank you for your hard work! Most of the changes are now resolved and the PR is on a good level. However, I notice a few spacing-related changes and only one scenario where the optimization does not work.

- ✅ Works when being uploaded from the `Media > Library` section.
- ❌ Does not work when uploaded from the `Media > Add New > (browser uploader; input field)` ← Also reloaded the page as per your comment but no luck.
- ✅ Works when uploaded from the `Media > Add New > (multi-file uploader; drag and drop)`
- ✅ Works with the core image block. (the post was saved before checking the size)
- ✅ Works with the core gallery block. (the post was saved before checking the size)
- ✅ Works with the Safe SVG block. (the post was saved before checking the size)
- ✅ No console error that was reported earlier.
- ✅ The filter `add_filter( 'safe_svg_svgo_params', '__return_false' );` works as expected.

---

> This is something that can be fixed, of course, but I will need to find some more time to do so.

Are you planning to work on fixing this (i.e. point no 2 from your comment above) in the current PR or a separate one?

gsarig:
> @gsarig thank you for your hard work! Most of the changes are now resolved and the PR is on a good level. However, I notice a few spacing-related changes and only one scenario where the optimization does not work.
> 
> * ✅ Works when being uploaded from the `Media > Library` section.
> * ❌ Does not work when uploaded from the `Media > Add New > (browser uploader; input field)` ← Also reloaded the page as per your comment but no luck.
> * ✅ Works when uploaded from the `Media > Add New > (multi-file uploader; drag and drop)`
> * ✅ Works with the core image block. (the post was saved before checking the size)
> * ✅ Works with the core gallery block. (the post was saved before checking the size)
> * ✅ Works with the Safe SVG block. (the post was saved before checking the size)
> * ✅ No console error that was reported earlier.
> * ✅ The filter `add_filter( 'safe_svg_svgo_params', '__return_false' );` works as expected.
> 
> > This is something that can be fixed, of course, but I will need to find some more time to do so.
> 
> Are you planning to work on fixing this (i.e. point no 2 from your comment above) in the current PR or a separate one?

Thanks for the quick reply @faisal-alvi I have some time this week so it's a good opportunity to finish this in the same PR

includes/optimizer.php
```diff
@@ -0,0 +1,160 @@
...
+				),
+				'before'
+			);
+		}
+		/**
+		 * Optimize the SVG file.
+		 *
+		 * @return void
+		 */
+		public function optimize() {
```
peterwilsoncc, line 120: This method needs a permission check (ie, `current_user_can( 'edit_post', $attachment_id )`).

It's a little confusing, but a `nonce` check only verifies that the request came from a logged in user, it doesn't check whether they have permission to perform the action.

assets/js/admin/admin.js
```diff
@@ -0,0 +1,300 @@
+import {optimize} from 'svgo/lib/svgo';
+import {select, subscribe} from '@wordpress/data';
+
+(function () {
+    const ajaxUrl = new URL(safeSvgParams.ajaxUrl); // eslint-disable-line no-undef
```
peterwilsoncc: Rather than disable the rule each time, I'd suggest defining the global in a comment at the start of the file

`/* global safeSvgParams */`

assets/js/admin/admin.js
```diff
@@ -0,0 +1,300 @@
...
+     * Optimize SVGs uploaded from the "Upload New Media" screen (`wp-admin/media-new.php`).
+     * @param {Element} form
+     */
+    const optimizeFromUploadNewMedia = (form) => {
+        if (form && form.classList.contains('html-uploader')) {
+            // We are on the browser uploader, so it's not possible to optimize on the fly. Instead, set a cookie to be checked after submission.
+            form.addEventListener('submit', function (event) {
+                const fileInput = document.querySelector('#async-upload');
+                if (fileInput.files[0].type === 'image/svg+xml' || fileInput.files[0].name.endsWith('.svg')) {
+                    document.cookie = safeSvgCookie + '=1;' + safeSvgCookieAttr;
```
peterwilsoncc: WordPress includes `wpCookies` with a bunch of cookie helper functions, it's enqueued on most pages in the admin so you should be able to use that.

peterwilsoncc (commented):
I've added a few notes inline.

While manually testing, I'm not seeing the changes been saved while logged in as an administrator.

I uploaded an SVG with a bunch of needless `<g>` tags, the optimiser picks it up client side and sends the network request, but the file isn't been updated.

includes/optimizer.php
gsarig: Thanks @peterwilsoncc I've added the check. As you can see, to get the `$attachment_id`, I used `attachment_url_to_postid( $svg_url )`.

assets/js/admin/admin.js
gsarig: @peterwilsoncc Done

assets/js/admin/admin.js
gsarig: Thanks for that tip @peterwilsoncc I've updated the code to use `wpCookies`. I couldn't find much documentation about `wpCookies` so I went to the core to see how it works. Can you please take a look to check if the implementation is good?

gsarig:
> I've added a few notes inline.
> 
> While manually testing, I'm not seeing the changes been saved while logged in as an administrator.
> 
> I uploaded an SVG with a bunch of needless `<g>` tags, the optimiser picks it up client side and sends the network request, but the file isn't been updated.

Thanks for the feedback @peterwilsoncc 
I've pushed fixes based on your feedback, and I've also merged `develop` into the branch, to solve some conflicts. 

I've been testing it with unoptimized SVGs taken [from here](https://unoptimized--svg-icon-stress-test.netlify.app/), as well as with the `custom.svg` from the Cypress tests, and on both cases I could see a difference in size between the unoptimized and the optimized file. 

One additional thing that I noticed was that Cypress reported a problem with a failed test: `Plugin should allow modify allowed tags and attributes:`. After some digging to understand why it happened, I realized that it sort of makes sense: When the optimizer runs, it cleans up any unknown tags and attributes that the devs might have passed using the `svg_allowed_attributes` and `svg_allowed_tags` filters. 

I searched in the SVGO library for any option to allow specific tags and attributes, but I couldn't find something. So, what I ended up doing was adding an additional check ([see here](https://github.com/10up/safe-svg/blob/57373727090ba156097932a931c84309f396a003/includes/optimizer.php#L52)) to disable the optimizer if a dev has added allowed tags or attributes. 

Please let me know what you think about that and if you have any other suggestions.

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+		 */
+		protected function url_to_path( string $url = '' ) {
+			if ( empty( $url ) ) {
+				return '';
+			}
+			$parsed_url = wp_parse_url( $url );
+			if ( empty( $parsed_url['path'] ) ) {
+				return false;
+			}
+			$file = ABSPATH . ltrim( $parsed_url['path'], '/' );
```
peterwilsoncc: I figured out why I wasn't getting the optimised version of the image.

I was testing on a site running WP in [sub-directory mode](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/) with the directory structure:
```
/ #root directory
/content #content directory
/wp # WordPress install.
```

Instead of `path/to/content/.../file.svg` the `$file` is calculated as `path/to/wp/content/.../file.svg`.

In the method above, I've suggested you use `get_attached_file()` which will remove the need for this method.

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+			);
+		}
+		/**
+		 * Optimize the SVG file.
+		 *
+		 * @return void
+		 */
+		public function optimize() {
+			$svg_url = filter_input( INPUT_GET, 'svg_url', FILTER_SANITIZE_URL );
+			if ( ! current_user_can( 'edit_posts', attachment_url_to_postid( $svg_url ) ) ) {
```
peterwilsoncc: Meta capabilities (ie, for the specific post) use the singular.

```suggestion
			if ( ! current_user_can( 'edit_post', attachment_url_to_postid( $svg_url ) ) ) {
```

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+				return;
+			}
+			check_ajax_referer( $this->nonce_name, 'svg_nonce' );
+			$svg_path = $this->url_to_path( $svg_url );
```
peterwilsoncc: You'll need to set `$attachment_id = attachment_url_to_postid( $svg_url );` above to avoid calling the function twice. It's quite expensive.

```suggestion
			$svg_path = get_attached_file( $attachment_id );
```

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+				'upload.php',
+				'media-new.php',
+			];
+			if ( ! in_array( $hook, $allowed_hooks, true ) ) {
+				return;
+			}
+			wp_enqueue_script(
+				'safe-svg-admin-scripts',
+				SAFE_SVG_PLUGIN_URL . '/dist/safe-svg-admin.js',
+				[ 'wp-data' ],
```
peterwilsoncc:
```suggestion
				[ 'wp-data', 'utils' ],
```

Just to make sure wpCookies is around. It will be for our purposes but best practice, etc, etc. :)

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+				SAFE_SVG_VERSION,
+				true
+			);
+			$params = wp_json_encode(
+				[
+					'ajaxUrl'    => esc_url( admin_url( 'admin-ajax.php' ) ),
```
peterwilsoncc: I'm unsure, but I a suspect this would be better as `sanitize_url()`, we're wanting to make sure the URL is safe -- json-encode makes sure it renders correctly for JS.

includes/optimizer.php
```diff
@@ -0,0 +1,162 @@
...
+					'svgoParams' => wp_json_encode( $this->svgo_params() ),
```
peterwilsoncc: I'm confused by this, why aren't you passing this as is to the outer wp_json_encode call -- then you won't need to call `JSON.parse` in the JavaScript file.

assets/js/admin/admin.js
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /**
+     * Hook into the WordPress Uploader and optimize the SVG.
+     */
+    if (wp.Uploader !== undefined) {
+        Object.assign(wp.Uploader.prototype, {
+            // Run on a successful upload.
+            success(attachment) {
+                const svgUrl = attachment?.attributes?.url;
```
peterwilsoncc, line 194: In the same data, you can get the ID of the image as `svgId = attachment?.attributes?.id` -- this may remove the need to call `attachment_url_to_postid` in the PHP, as it's quite an expensive function.

peterwilsoncc (commented):
I've added a few more notes inline, including why the optimization was failing when I tested earlier.

In the JS, it should be possible to use the attachments ID and pass that to your ajax request in place of the URL. That will make the admin-ajax function a little more straight forward.

assets/js/admin/admin.js
gsarig: Done.

includes/optimizer.php
gsarig: Fixed

includes/optimizer.php
gsarig: `sanitize_url()` has been deprecated, and WP suggests using `esc_url_raw()` instead, so that's what I changed it to.

includes/optimizer.php
gsarig: Done!

includes/optimizer.php
gsarig: Done.

includes/optimizer.php
gsarig: You are right, I missed that. This should be fixed now.

includes/optimizer.php
gsarig: Thanks for that @peterwilsoncc I followed your suggestion and on my local tests it seems to be working.

gsarig:
> I've added a few more notes inline, including why the optimization was failing when I tested earlier.
> 
> In the JS, it should be possible to use the attachments ID and pass that to your ajax request in place of the URL. That will make the admin-ajax function a little more straight forward.

Thank you for the great feedback @peterwilsoncc 
I pushed some fixes based on your suggestions.

peterwilsoncc (approved):
LGTM, thanks for your patience during the review.

I noticed the optimize ajax request was been made twice on each save when I uploaded directly in to the block editor but have not been able to figure out why. It appears the `wp.data` subscription runs twice but that doesn't seem like something we have control over.

I think this is good to merge but attach a gif of what I am seeing in case inspiration strikes you.

double-optimize

darylldoyle:
> One additional thing that I noticed was that Cypress reported a problem with a failed test: `Plugin should allow modify allowed tags and attributes:`. After some digging to understand why it happened, I realized that it sort of makes sense: When the optimizer runs, it cleans up any unknown tags and attributes that the devs might have passed using the `svg_allowed_attributes` and `svg_allowed_tags` filters. 

I'd love to know what sort of attributes we're seeing stripped here. This is something I was looking into with the previous Pro implementation of SVGO and I have a feeling there are ways around it using SVGO plugins

For example, this is the plugin that removes the unknowns. The knowns are stored in a collections file. Perhaps we could extend that with our custom properties 

https://github.com/svg/svgo/blob/main/plugins/removeUnknownsAndDefaults.js

PR #80, opened 2022-10-12
Feature/21 An SVG Gutenberg Block

### LINKED ISSUE #21: Consider adding an inline SVG Gutenberg block
**Is your enhancement related to a problem? Please describe.**

Attempting to add an SVG onto a page/post with the image block presents all sorts of issues. The core image block expects images to have an intrinsic size even without width & height attributes, as the block uses an extra wrapper div to set the width/height in the editor. This situation leads to all sorts of weirdness with the size of the SVG in the editor vs the front end, as well as often making the image disappear if you select an alignment, or even just remove focus from the block.

Image block with no settings changed, no alignment, selected:
Screen Shot 2022-01-13 at 3 36 08 PM

Same as above, block no longer selected:
Screen Shot 2022-01-13 at 3 36 47 PM

Block set to align center makes image 0 width again:
Screen Shot 2022-01-13 at 3 38 18 PM

On the front end, with no alignment set, the width and height attributes are added, which is a different width than the default width while block is selected from screenshot number 1 above.
Screen Shot 2022-01-13 at 3 39 47 PM

**Describe the solution you'd like**

As the image block seems to be fundamentally unable to handle SVG images, and only with this plugin can you upload SVGs anyways, I propose that the plugin include a custom block for outputting an SVG markup directly.

In places that aren't the block editor, we've used this plugin to allow for SVG uploads and then output the markup in theme code, but we don't have a way of doing that without creating a custom block anyways.

**Designs**

**Describe alternatives you've considered**

If it's possible to add support for handling the SVG in the core image block better, perhaps that would be a better route, but I'm not sure if that is feasible and maintainable.

**Additional context**

faisal-alvi:
### Description of the Change

* This PR introduces a new SVG block.
* Inspired by the core image block but it only allows SVG files to be selected/uploaded.
* Adds "Alignment" and "Replace" Block controls the same as the core image block.
* Also adds the Image settings in the block sidebar (image size dropdown, image dimensions, with a reset button).

image
[image link](https://www.screencast.com/t/aZImKzQLDPy)

image
[image link](https://www.screencast.com/t/5w0MkoCPA)

image
[image link](https://www.screencast.com/t/TdhRqYMzFM8)

image
[image link](https://www.screencast.com/t/FG6WB4Rx)

Closes #21

### How to test the Change

1. Create a new page/post.
2. Add a block "Safe SVG Icon".
3. Click "Select an SVG icon".
4. Try alignment, replace, various image sizes.
5. Confirm the FE should be same as BE.

### Credits

Props @faisal-alvi @jeffpaul @Sidsector9 

### Checklist:


Sidsector9 (changes requested):
@faisal-alvi thanks for working on this.

I have used this image to test:

wide-svg


Few things I would like to point out:

1. When you first select the SVG, the rendered image dimensions is very small, even when **Image size** is set to **Full Size**.
2. When you first select the SVG, the default values set are: Full Size, 200x200, 100%, but when you switch to Medium and back to Full Size, the dimension changes to 575x188.
3. The SVG width doesn't stretch to 100% width of the editor when setting dimension to 100%.
4. Without an SVG selected, the controls are practically dead. We can hide settings if the SVG is not selected.

Please watch the video for more details, I have compared it with the `core/image` block.

https://user-images.githubusercontent.com/17757960/196629795-41df8dbe-c091-4c3e-ae8b-b4ec6fb82c70.mp4

## Note

PR #79 uses 10up-toolkit whereas this one uses wp-scripts. We need to decide which one to go ahead with and make changes to the other PR accordingly.

darylldoyle:
@jeffpaul @faisal-alvi I have to agree with @cr0ybot here. The reason the original pro version of the plugin used a dynamic block, was to get around the issue with the `unfiltered_html` capability. Especially when it comes to WordPress multisite where only super admins have that capability out of the box.

We're already restricting _who_ can upload in #76 and unless we also want to give those roles the `unfiltered_html` capability (we don't), then this will become a sticking point of the block.

I suggest we stick to a dynamically rendered block. 

Original implementation for reference: https://github.com/10up/wpsvg/blob/develop/gutenberg/class-wpsvg-gutenberg.php#L103-L194

faisal-alvi:
@cr0ybot @darylldoyle thank you for the feedback, it all makes sense. I will go ahead and tweak the code to make the block dynamically rendered at the front end. I assume that we have to keep the `viewbox` value as it is and keep whatever we receive from the `file_get_contents`, and only apply the height and width `$attr`s to the `<svg>` image at the front end. Please let me know if this approach is fine.

faisal-alvi:
@Sidsector9 thanks for the feedback.

> When you first select the SVG, the rendered image dimensions is very small, even when Image size is set to Full Size.

done

> When you first select the SVG, the default values set are: Full Size, 200x200, 100%, but when you switch to Medium and back to Full Size, the dimension changes to 575x188.

done

> The SVG width doesn't stretch to 100% width of the editor when setting dimension to 100%.

I don't know if that was an issue, but I've checked after the recent changes and we can set the SVG image to cover 100% of the available width.

> Without an SVG selected, the controls are practically dead. We can hide settings if the SVG is not selected.

done

jeffpaul:
@darylldoyle @Sidsector9 any further code review feedback here or is this good to move along to merge/release?

jeffpaul:
@faisal-alvi note some merge conflicts to get cleaned up here, please and thanks!

darylldoyle (approved):

darylldoyle:
LGTM! Nice work 🙂

ocean90:
Some quick feedback for this new block:

* Why the `imageSizes` attribute which just blows up the block content. A SVG image only has one size?
  ```html
   <!-- wp:safe-svg/svg-icon {"svgURL":"https://example.com/content/uploads/2022/07/desktop-tab-icon.svg","type":"medium","alignment":"center","imageID":36100,"imageWidth":300,"imageHeight":300,"dimensionWidth":300,"dimensionHeight":300,"imageSizes":{"full":{"height":48,"width":48,"url":"https://example.com/content/uploads/2022/07/desktop-tab-icon.svg","orientation":"portrait"},"medium":{"height":"300","width":"300","url":"https://example.com/content/uploads/2022/07/desktop-tab-icon.svg","orientation":"portrait"},"thumbnail":{"height":"150","width":"150","url":"https://example.com/content/uploads/2022/07/desktop-tab- icon.svg","orientation":"portrait"}}} /-->
  ```
* All pages now load the block stylesheet even if no block is used on the page
* There's a double slash in the URL: https://example.com/content/plugins/safe-svg//dist/safe-svg-block-frontend.css?ver=2.1.0
* `init` is an action, [not a filter](https://github.com/10up/safe-svg/blob/acbac59fed1b6a274f0cfc7fd00f961218685f6e/safe-svg.php#L81)

---

In case anyone is looking for disabling the new block:

```php
// Disable Safe SVG block.
add_action(
	'plugins_loaded',
	static function(): void {
		global $safe_svg;

		if ( ! isset( $safe_svg ) ) {
			return;
		}

		remove_action( 'init', [ $safe_svg, 'setup_blocks' ] );
	}
);
```

PR #112, opened 2023-03-29
Fix issues with how assets are enqueued on the new SVG block

### LINKED ISSUE #109: Safe-SVG v2.1.0 should not enqueue frontend stylesheet and JavaScript if Gutenberg block is not used

The new version 2.1.0, which adds a Gutenberg block, enqueues the following files in the frontend of the website, even if this block is not used:
* safe-svg-block-frontend.css
* safe-svg-block-frontend.js

This is also mentioned in https://github.com/10up/safe-svg/pull/80#issuecomment-1482534927

The JS file is even empty (!). As a web agency focused on the performance of our websites, we keep an eye on every single CSS and JS file loaded in the frontend. Please do not enqueue CSS and JS files for unused functionality. Especially in a plugin, which previously only provided backend functionality (and we will still only use it for that).

Right now we have to add the following code to all of our websites using this plugin to dequeue these assets:
```php
wp_dequeue_style('safe-svg-block-frontend');
wp_deregister_script('safe-svg-block-script');
```

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

A new SVG block was added in PR #80, which went out in the 2.1.0 release. We've gotten a handful of reports around how assets are being loaded for this block, most notably:

- a CSS and JS file are loaded on all FE pages, even those without the block
- the JS file is empty
- an extra slash is in the CSS file URL

This PR fixes all of those issues by removing the JS file from being loaded at all and moving the CSS file from a typical enqueue to being loaded as part of the `block.json`. This ensures the CSS is only loaded on pages that are using the block and it loads the CSS inline instead of loading an extra stylesheet.

Also fix an incorrect use of `add_filter` to be `add_action`

Closes #109 

### How to test the Change

On the current released version (2.1.0) view any page on your site and note that two block related files (a CSS and JS file) are being loaded.

Checkout this branch and note those files are no longer loaded at all, even if a page has an SVG block. Verify the block itself still works and renders as expected.

### Changelog Entry

> Fixed - Only load our block CSS if a page has the SVG block in it and remove an extra slash in the CSS file path.. Remove an unneeded JS block file.

### Credits

Props @dkotter, @freinbichler, @IanDelMar, @ocean90

### Checklist:

### DIFF
--- a/includes/blocks.php
+++ b/includes/blocks.php
@@ -17,7 +17,6 @@ function setup() {
 		return __NAMESPACE__ . "\\$function";
 	};
 
-	add_action( 'enqueue_block_assets', $n( 'blocks_styles' ) );
 	add_filter( 'block_categories_all', $n( 'blocks_categories' ), 10, 2 );
 
 	register_blocks();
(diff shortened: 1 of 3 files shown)


Sidsector9 (approved):
LGTM 👍

PR #135, opened 2023-07-17
Improved UX with spacing & color options

### LINKED ISSUE #131: Safe SVG Icon Block UX Suggestions
Whilst working on a site that is using the Plugin, I interacted with the Safe SVG Icon block for the first time and I have some thoughts that could improve the usability :) 

**1. Change the name and or description of the block to be more descriptive**

When I first saw the block, it took me quite a while to figure out why I would want to use the block over just using an Image block to select the SVG. The reason why you would want to use the block over the core image is that it inlines the SVG. So my suggestion is to rename the block to `Inline SVG` or at least add something to the block description and keywords about it rendering the SVG inline. 

**2. Change the block Icon**

Currently the Icon doesn't feel very at home in the editor. It is to bold / weighty compared to other core blocks. I would suggest replacing it with an icon that is more similar to the icon of the Custom HTML block that just says `SVG` or something similar :) 
CleanShot 2023-06-22 at 13 24 46@2x

CleanShot 2023-06-22 at 15 22 15@2x

**3. Refine the "empty" state**

Currently the empty / placeholder state of the block only renders a button which opens the media modal. 
CleanShot 2023-06-22 at 15 18 59@2x

This is very different to how all the other media blocks handle this. The image block for example renders a `MediaPlaceholder` which also allows you do drop something into the placeholder directly to upload it etc. 
CleanShot 2023-06-22 at 15 20 29@2x

I think we can do something more similar to that very easily by making use of the [`MediaPlaceholder`](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-editor/src/components/media-placeholder) component

**4. Add color controls**

Since the block actually renders an SVG element we have the benefit of being able to style the SVG. One of the most common things we need to do is to be able to change the color of a solid icon. I think it would be great to add the `text` color support so that an SVG that uses the `currentColor` value is able to be modified via the block.

**5. Add dimension controls**

Similar to the image block it would be great to be able to change the size of an icon. And whilst we are at it being able to control margin and padding are also great options to have for this kind of element. (If we add padding we should also add background color support :))

bmarshall511:
### Description of the Change

1. Updated the name of the block to be clearer in what it provides
2. Added spacing controls
3. Added BG & text color controls
4. Replaced the button to select media with the `MediaPlaceholder` component
5. Updated the block icon

Closes #131 

### How to test the Change

1. Update/create a post/page
2. Add an `Inline SVG` block
3. Use an SVG that has a `fill` of `currentColor`
4. Select a text color to change the SVG fill
6. Select a background color to change the background
7. Test different alignments (left, center, right)

### Changelog Entry

> Added - Added Spacing & color controls & a more descriptive block name

### Credits

Props @bmarshall511 

### Checklist:


includes/blocks/safe-svg/edit.js
```diff
@@ -155,47 +157,31 @@ const SafeSvgBlockEdit = ( props ) => {
...
+					value={imageID}
+					labels={{
+						title: __( 'Inline SVG', 'safe-svg' ),
+						instructions: __( 'Upload an SVG or pick one from your media library.', 'safe-svg' )
+					}}
+				/>
+			}
+
+			{svgURL &&
+				<div { ...containerBlockProps } className={className.replace(/(has-vivid-red-background-color|has-secondary-color|has-text-color|has-background)/g, '').trim()}>
```
iamdharmesh: Maybe replace other background colors as well?
```suggestion
				<div { ...containerBlockProps } className={className.replace(/(has-vivid-.*-background-color|has-secondary-color|has-text-color|has-background)/g, '').trim()}>
```

includes/blocks/safe-svg/register.php
```diff
@@ -67,18 +67,44 @@ function render_block_callback( $attributes ) {
 	return apply_filters(
 		'safe_svg_inline_markup',
 		sprintf(
-			'<div class="safe-svg-cover" style="text-align:%s">
-				<div class="safe-svg-inside %s%s" style="width: %spx; height: %spx;">%s</div>
+			'<div class="safe-svg-cover align%s">
```
iamdharmesh: Maybe we should keep the alignment behavior as earlier? I have no issue with new alignment updates but it may break layouts on site which are using this plugin. what do you think?

| Develop Branch | This PR |
| --- | --- |
|Screenshot 2023-07-20 at 12 26 51 PM|Screenshot 2023-07-20 at 12 27 51 PM|
|Screenshot 2023-07-20 at 12 27 01 PM|Screenshot 2023-07-20 at 12 28 04 PM|

iamdharmesh (changes requested):
Thanks for the amazing work here @bmarshall511. The code looks good to me and it tests very well. Just added 2 minor notes to discuss otherwise, all is good here.

Thanks

includes/blocks/safe-svg/edit.js
bmarshall511: @iamdharmesh Yeah, great catch! Updated the regex so it removes all classes that start with `has-` and ends with `-color` along with the existing `has-background` class.

includes/blocks/safe-svg/register.php
bmarshall511: @iamdharmesh Good point! Was trying to implement how the `core/image` block works since this is essentially the same, expect it being an inlined SVG. I've updated it so works as it did before.

includes/blocks/safe-svg/edit.js
```diff
@@ -155,47 +174,34 @@ const SafeSvgBlockEdit = ( props ) => {
...
+						title: __( 'Inline SVG', 'safe-svg' ),
+						instructions: __( 'Upload an SVG or pick one from your media library.', 'safe-svg' )
+					}}
+				/>
+			}
+
+			{svgURL &&
+				<div className={newClassName} style={{ textAlign: alignment }}>
+					<div
+						{ ...containerBlockProps }
```
iamdharmesh: Thanks for making the changes @bmarshall511. I think we should move this to the root `<div>` element. Otherwise, the user will not be able to select a block by clicking on it.

bmarshall511: @iamdharmesh Ah, nice catch! Sorry about that. I've updated it accordingly.

iamdharmesh (approved):
Thanks for the changes @bmarshall511. It looks great now. 🚀

PR #137, opened 2023-07-20
Added a helper function for wp_kses to sanitize a SVG

### LINKED ISSUE #115: It would be beneficial if you could use `safe-svg` for `wp_kses` and other similar purposes.

I am using `svg` in my widget (custom HTML block) and sanitizing the content with `wp_kses`. However, it strips down the svg markup.

It would be nice if `safe-svg` could provide a helper function for such use cases. It wouldn't necessarily need to hook into `wp_kses_allowed_html`, but it would be helpful if I could use the sane defaults provided by this plugin.

`wp_kses( $input, \enshrined\svgSanitize\Sanitizer\kses_extended_ruleset() )` or such would be the use.

### Designs

_No response_

_No response_

### Code of Conduct

bmarshall511:
### Description of the Change

Added a helper function used for `wp_kses()` to return an array of allowed HTML tags & attributes.

Closes #115

### How to test the Change

Example usage:

```php
echo wp_kses('<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 25"><path fill="currentcolor" d="M17.525 9.302H14v-2c0-1.032.084-1.682 1.563-1.682h1.868V2.44a26.065 26.065 0 0 0-2.738-.138C11.98 2.302 10 3.959 10 7v2.3H7v4h3v9h4V13.3l3.066-.001.459-3.996Z"/></svg>', \SafeSvg\SafeSvgTags\safe_svg_tags::kses_allowed_html())
```

### Changelog Entry

> Added - Added the `kses_allowed_html` helper function for `wp_kses` to sanitize SVG images

### Credits

Props @bmarshall511 

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,14 @@ add_filter( 'svg_allowed_tags', function ( $tags ) {
 } );
 ```
 
+### Can `wp_kses` be used with a helper to sanitize an SVG?
+
+Indeed, you can accomplish this with `\SafeSvg\SafeSvgTags\safe_svg_tags::kses_allowed_html()`:
+
+```php
+echo wp_kses('<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 25"><path fill="currentcolor" d="M17.525 9.302H14v-2c0-1.032.084-1.682 1.563-1.682h1.868V2.44a26.065 26.065 0 0 0-2.738-.138C11.98 2.302 10 3.959 10 7v2.3H7v4h3v9h4V13.3l3.066-.001.459-3.996Z"/></svg>', \SafeSvg\SafeSvgTags\safe_svg_tags::kses_allowed_html())
+```
+
 ## Support Level
 
 **Stable:** 10up is not planning to develop any new features for this, but will still respond to bug reports and security concerns. We welcome PRs, but any that include new features should be small and easy to integrate and should not include breaking changes. We otherwise intend to keep this tested up to the most recent version of WordPress.
--- a/tests/unit/test-safe-svg-tags.php
+++ b/tests/unit/test-safe-svg-tags.php
@@ -47,4 +47,14 @@ class SafeSvgTagsTest extends TestCase {
 		$this->assertContains( 'customTag', $svg_tags );
 		$this->assertSame( $svg_tags, $filtered_svg_tags );
 	}
+
+	/**
+	 * Test the kses_allowed_html function.
+	 *
+	 * @throws PHPUnit\Framework\AssertionFailedError If the function does not return an array.
+	 */
+	public function test_kses_allowed_html() {
+		$allowed_html = SafeSvg\SafeSvgTags\safe_svg_tags::kses_allowed_html();
+		$this->assertIsArray( $allowed_html );
+	}
 }
--- a/includes/safe-svg-tags.php
+++ b/includes/safe-svg-tags.php
@@ -24,4 +24,98 @@ class safe_svg_tags extends \enshrined\svgSanitize\data\AllowedTags {
 		 */
 		return apply_filters( 'svg_allowed_tags', parent::getTags() );
 	}
+
+	/**
+	 * Standard SVG settings for escaping through `wp_kses()` function.
+	 *
+	 * @return array Array of allowed HTML tags and their allowed attributes.
+	 */
+	public function kses_allowed_html() {
+		return array(
+			'svg'            => array(
+				'version'           => true,
+				'class'             => true,
+				'fill'              => true,
+				'height'            => true,
+				'xml:space'         => true,
+				'xmlns'             => true,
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -74,6 +74,14 @@ add_filter( 'svg_allowed_tags', function ( $tags ) {
 } );
 ```
 
+### Can `wp_kses` be used with a helper to sanitize an SVG?
```
darylldoyle, line 77: **issue:** kses will not sanitise SVGs, so we either need to change this title, or re-think the use-case for this filter.

tests/unit/test-safe-svg-tags.php
```diff
@@ -47,4 +47,14 @@ public function test_get_tags() {
...
 	}
+
+	/**
+	 * Test the kses_allowed_html function.
+	 *
+	 * @throws PHPUnit\Framework\AssertionFailedError If the function does not return an array.
+	 */
+	public function test_kses_allowed_html() {
+		$allowed_html = SafeSvg\SafeSvgTags\safe_svg_tags::kses_allowed_html();
+		$this->assertIsArray( $allowed_html );
```
darylldoyle, line 58: **thought (non-blocking):** This feels like adding a test just to have a test. I'd prefer it if we were either verifying that the correct items were stripped from an SVG, or that the correct params/attributes are included.

darylldoyle (changes requested):
I have a few concerns with this PR.

1. I'm not sure what this solves. KSES cannot be used as an SVG sanitisation method, see [here](https://core.trac.wordpress.org/ticket/24251#comment:73). Is it just to help with output? If so, why not use `wp_kses_post()`?
2. There is a list of SVG elements for KSES but no listed attributes which means we'll allow them all. Within the SVG specification, the SVG attributes pose more of a threat to an XSS vulnerability than the elements do.
3. To keep this in line with the `svg-sanitiser` library, this list will need to be kept up to date with each release of the underlying library. This will likely be labour-intensive.
4. The Safe SVG plugin includes filters that allow engineers to extend the list of allowed elements <sup>[1](https://github.com/10up/safe-svg/blob/8e579a74076ffc35add787dd5eb328f588fcea35/includes/safe-svg-tags.php#L25)</sup> and attributes <sup>[2](https://github.com/10up/safe-svg/blob/8e579a74076ffc35add787dd5eb328f588fcea35/includes/safe-svg-attributes.php#L25)</sup>. This KSES filter doesn't take that into account and therefore could cause the allowed elements/attributes to be stripped by late escaping.

Whilst I think it's great to push this plugin forward, I think we need to consider how we can ensure we're not overcomplicating things and to always be cognizant of any potential security impacts.

If you're interested in further reading, the below links might be interesting:

1. [The core trac ticket to get SVG support added](https://core.trac.wordpress.org/ticket/24251)
2. [An old blog post from me on SVG security](https://enshrined.co.uk/2018/04/29/securing-svg-uploads-in-wordpress/)
3. [My WordCamp London 2018 talk about SVG security](https://wordpress.tv/2019/01/22/daryll-doyle-securing-svg-uploads-in-wordpress/)
4. [Mario Heiderich's, The Image that called me presentation](https://owasp.org/www-chapter-oslo/assets/files/The_image_that_called_me.pdf)

bmarshall511:
@darylldoyle Totally get where you're coming from. This was actually in response to this issue: https://github.com/10up/safe-svg/issues/115. In a few projects that I've been dabbling in recently, we've been using something similar within the theme to sanitize SVG output, so thought it might be a pretty cool solution for that request.

You've made some solid points, especially about KSES not being suitable for SVG sanitization. And I agree, maintaining the SVG elements list in line with the svg-sanitiser library might be quite a handful. 

I hear ya about the fact that the KSES filter doesn't consider the extendibility of the list of allowed elements/attributes. This could indeed lead to some elements/attributes being stripped off due to late escaping. And for sure, the last thing we want is to overcomplicate things and neglect potential security implications.

Gonna check out those resources you've shared - seems like some really useful stuff there. I'm all about that SVG security knowledge, so much appreciated for that! How about for now I close this & can be used as reference for the #115 issue?

darylldoyle:
@bmarshall511 I appreciate your willingness to understand this! I hope the shared resources are helpful 🙂 

In regards to #115, I think it'd be useful for Safe SVG to provide a helper function which wraps these lines:

https://github.com/10up/safe-svg/blob/develop/safe-svg.php#L277-L283

That would give people a way to output SVGs whilst ensuring they're sanitised. There's something to say for sanitising it to appease a linter. If the SVG is bundled with a theme, do we need to sanitise it on output? We should know what's in that file. We mainly need to sanitise files coming from third parties.

It would make sense if we were outputting SVGs from an API or other third-party integration.

I'm happy for you to close this issue and move the discussion to #115 if you like.

PR #140, opened 2023-08-01
Fix crash if SimpleXML is missing

### LINKED ISSUE #139: Broken media management if simplexml is missing
### Describe the bug

When an SVG is uploaded, the plugin calls `simplexml_load_file` to retrieve the SVG dimensions from the file. However, if the `simplexml` PHP extension is not installed, the plugin will crash while attempting to generate the metadata for the image.

If you use the media gallery to upload a file then the upload process won't complete and WordPress will throw a generic error message:

> The server cannot process the image. This can happen if the server is busy or does not have enough resources to complete the task. Uploading a smaller image may help. Suggested maximum size is 2560 pixels.

However if you upload an image via the REST API, the request will fail with a 500 HTTP error but the image will still be successfully uploaded without metadata. This becomes a problem because you won't be able to access the media gallery afterwards since the plugin will attempt to regenerate the metadata on the fly and crash.

### Steps to Reproduce

1. Upload an image via the REST API. Request will fail with a 500 error.
2. File will exist in the uploads folder and in the `wp_posts` table, but there won't be a `_wp_attachment_metadata` in `wp_postmeta`.
3. Try to access the media gallery.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

sdmtt:
Fixes 10up/safe-svg#139

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -564,6 +564,10 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 		 * @return array|bool
 		 */
 		protected function svg_dimensions( $svg ) {
+			if ( ! function_exists('simplexml_load_file') ) {
+				return false;
+			}
+
 			$svg    = @simplexml_load_file( $svg ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
 			$width  = 0;
 			$height = 0;
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -564,7 +564,7 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 		 * @return array|bool
 		 */
 		protected function svg_dimensions( $svg ) {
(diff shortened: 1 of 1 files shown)


faisal-alvi (approved):
Tested, and works as expected! Thank you @sdmtt

safe-svg.php
```diff
@@ -564,6 +564,10 @@ public function metadata_error_fix( $data, $post_id ) {
 		 * @return array|bool
 		 */
 		protected function svg_dimensions( $svg ) {
+			if ( ! function_exists('simplexml_load_file') ) {
```
faisal-alvi:
```suggestion
			if ( ! function_exists( 'simplexml_load_file' ) ) {
```

faisal-alvi (changes requested):
a minor spacing issue

faisal-alvi (approved):

PR #145, opened 2023-08-16
Disable SVG optimization by default

dkotter:
### Description of the Change

In #79, we added SVG optimization using the SVGO library. This optimization is enabled by default and the only way to turn it off was through a filter. I had some concerns with this feature (around performance and potential issues with stripping params we may want to keep) and so in order to not delay a 2.2.0 release any longer, talked to @jeffpaul and decided that we would keep this feature in place but we would change this to be disabled by default.

A new filter has been added (`safe_svg_optimizer_enabled`) that returns `false` by default. Users that want to try out this new optimization feature will have to hook into that filter and return `true`:

```php
add_filter( 'safe_svg_optimizer_enabled', '__return_true' );
```

This gives people the opportunity to test this out and decide whether they want this extra optimization or not (and doesn't impact sites that don't care about this). If enough people use and like this feature, in the future we can look to either enable this by default or add an admin setting to turn this on/off.

### How to test the Change

Follow the steps outlined in #79, once you add the line of code above. Without that line in place, ensure optimization doesn't happen and everything works as expected.

### Changelog Entry

> Changed - Don't run SVGO optimization by default and instead make this an opt-in feature.

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/optimizer.php
+++ b/includes/optimizer.php
@@ -10,22 +10,26 @@ namespace SafeSVG;
 use enshrined\svgSanitize\Sanitizer;
 
 if ( ! class_exists( '\SafeSVG\Optimizer' ) ) {
+
 	/**
 	 * Class \SafeSVG\Optimizer
 	 */
 	class Optimizer {
+
 		/**
 		 * The name of the nonce to send with the AJAX call.
 		 *
 		 * @var string
 		 */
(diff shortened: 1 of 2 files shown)


ravinderk (approved):
@dkotter It is working as expected.

PR #152, opened 2023-09-15
Fixed fill-rule browser error

### LINKED ISSUE #149: Browser Warning Error for svg icon
### Describe the bug

After installing this plugin and opening the block inserter, the following error is output to the browser console:

```
react-dom.js?ver=18.2.0:73 Warning: Invalid DOM property `fill-rule`. Did you mean `fillRule`?
```

This SVG is rendered as a React element, so I think it needs to be changed to `fillRule`

https://github.com/10up/safe-svg/blob/0e895dadd327df9849f827d858675e9be8b948ef/includes/blocks/safe-svg/index.js#L33

### Steps to Reproduce

You may need to change `SCRIPT_DEBUG` to `true` to reproduce this problem.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

zamanq:
### Description of the Change
Fixed the browser error for svg icon that was caused by the `fill-rule` attribute.
Closes #149 

### How to test the Change
=> In `wp-config.php` file, enable script debug: `define( 'SCRIPT_DEBUG', true );`
=> Create a new page and add the Inline SVG block from block inserter
=> Add any svg image and the browser warning should go away.

### Credits
Props @zamanq @jeffpaul 

### Checklist:


iamdharmesh (approved):
Thanks for the PR @zamanq. Code LGTM and it tests well.

PR #154, opened 2023-09-25
Fix | Slow load times due to svg_dimensions

### LINKED ISSUE #75: Slow load times due to svg_dimensions
### Describe the bug

Some time ago I opened [this issue](https://wordpress.org/support/topic/slow-performance-of-svg_dimensions-function/) regarding the performance of this plugin (also see [this related issue](https://github.com/LF-Engineering/lfevents/issues/546)). Basically when we have a lot of SVGs on a page populated by a Gutenberg block callback function, the load time for the page can be very long due to multiple calls to the svg_dimensions function. At the time the issue was not resolved so I edited the plugin [as described](https://wordpress.org/support/topic/slow-performance-of-svg_dimensions-function/#post-13215804) to avoid overuse of the svg_dimensions function.

Now that this plugin is being developed again, could it be modified to include a filter to avoid this operation. [You can see how I created the filter in my edited version of the plugin here](https://github.com/LF-Engineering/lfevents/commit/b31835692a5b79b0431f4a7b0f8957d284dee32e). Alternatively, the performance issues could be addressed in some other way. Then I could stop using my custom version.

### Steps to Reproduce

Create a page with a ton of SVGs that get populated using a custom Gutenberg block callback function. [See this comment for more info](https://wordpress.org/support/topic/slow-performance-of-svg_dimensions-function/#post-13103829).

See [this image](https://user-images.githubusercontent.com/463160/82769273-6796a480-9e66-11ea-96eb-e535a15f8caf.png)

### Environment information

not relevant

### WordPress information

This bug was reported 2 years ago so I was on an older version of WP then but I doubt the WP version is relevant.

### Code of Conduct

### LINKED ISSUE #151: Function used by Save SVG is slowing down wordpress pages
### Describe the bug

Currently, I use a theme in my WordPress sites that uses the function wp_get_attachment_image for rendering image tags. This function takes the attachment ID and renders its HTML tag (``).

I have a huge database with +4,000,000 rows in my wp_posts table. The function `wp_attachment_get_image` will take 0.1 to 0.2 seconds for rendering SVGs, while the Jpegs and PNGs take only 0.0001 seconds for rending the same tag with the same functions. 

Turning on the PHP slow execution log, I could trace back the cause of the slowness. It is the function [simplexml_load_file](https://www.php.net/manual/pt_BR/function.simplexml-load-file.php). I added below the file that is using it:
image

So, to summarize, those are the times I have when the Safe SVG is turned on:
image

And those are the times I have with it off:
image

Your plugin is by far the most used by my company. I really would appreciate it if you help me to solve the issue.

### Steps to Reproduce

For tracking the timing I used the plugin Query Monitor, together with the following hooks around the function wp_get_attachment_image for tracking the execution time:
```
do_action( 'qm/start',$this->img_id);
$image = wp_get_attachment_image( $this->img_id, $this->size, false, $this->attr );
do_action( 'qm/stop',$this->img_id);
```

Another thing you could do in order to reproduce the error would be just to render 30 to 40 SVGs on a page. It would slow the page down, consuming 4 to 10 seconds for rendering the page.

_No response_

### Environment information

PHP 8.2
WordPress 6.3
Safe SVG 2.2.0

### WordPress information

_No response_

### Code of Conduct

sksaju:
### Description of the Change

We have refactored the `svg_dimensions` function to calculate dimensions using the existing attachment meta.

Closes [#75](https://github.com/10up/safe-svg/issues/75) and [#151](https://github.com/10up/safe-svg/issues/151)

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:


Sidsector9:
@sksaju the PR is in draft, is this ready for review?

sksaju:
> @sksaju the PR is in draft, is this ready for review?

Not yet. I just need to fix the unit test issue, which I can do early next week :)

safe-svg.php
```diff
@@ -627,11 +634,21 @@ protected function svg_dimensions( $svg ) {
...
+
+			/**
+			 * Calculate SVG dimensions and orientation.
+			 *
+			 * @param array  $dimensions An array containing width, height, and orientation.
+			 * @param string $svg        The file path to the SVG.
+			 *
+			 * @return array An array of SVG dimensions and orientation.
+			 */
+			return apply_filters( 'safe_svg_dimensions', $dimensions, $svg );
```
darylldoyle, line 668: @sksaju @10up/open-source-practice I like this filter, but what are your thoughts on introducing a `safe_svg_pre_dimensions` filter also? 

My thinking is, if someone wants to override this value, forcing them through the overhead of loading and sizing the SVG seems a little odd.

Something like the following at the top of this method to allow the short-circuit might be useful:

```php
/**
* Calculate SVG dimensions and orientation.
*
* This filter allows you to implement your own sizing. By returning a non-false
* value, it will short-circuit this function and return your set value.
*
* @param array  $attachment_id The attachment ID of the SVG being processed.
*
* @return array|false An array of SVG dimensions and orientation or false.
*/
$short_circuit = apply_filters( 'safe_svg_pre_dimensions', '__return_false' );

if ( false !== $short_circuit ) {
    return $short_circuit;
}

```

sksaju: Thanks @darylldoyle I've updated the PR, please take a look again and let me know your feedback. could you also let me know how to fix this https://github.com/10up/safe-svg/actions/runs/7004501147/job/19052370329?pr=154 UnitTest issue? Thanks

jeffpaul:
@darylldoyle back to you for review

safe-svg.php
```diff
@@ -560,19 +560,43 @@ public function metadata_error_fix( $data, $post_id ) {
...
+			 * @param boolean Default value of the filter.
+			 * @param integer $attachment_id The attachment ID of the SVG being processed.
+			 *
+			 * @return array|false An array of SVG dimensions and orientation or false.
+			 */
+			$short_circuit = apply_filters( 'safe_svg_pre_dimensions', false, $attachment_id );
+
+			if ( false !== $short_circuit ) {
+				return $short_circuit;
+			}
```
darylldoyle, line 583: **praise (non-blocking):** Nice work here!

darylldoyle (dismissed):
This looks great @sksaju! 

The PHPUnit failure looks like it's because you haven't mocked `wp_get_attachment_metadata()` within the test. I didn't _think_ it would be needed, but worth a try. See: https://github.com/10up/safe-svg/blob/develop/tests/unit/test-safe-svg.php#L191-L209

Once those tests are passing I'm good with merging this. 

CC @jeffpaul

darylldoyle (dismissed):
Looking good! Ignoring E2E tests as it doesn't look like they're fully set-up

sksaju:
> Looking good! Ignoring E2E tests as it doesn't look like they're fully set-up

Thanks @darylldoyle I already fixed the E@E issue here https://github.com/10up/safe-svg/pull/154/commits/d0b5b8a2a7d9776bfa639f9b912ccf090686e01c :)

darylldoyle (approved):

safe-svg.php
```diff
@@ -560,19 +560,43 @@ public function metadata_error_fix( $data, $post_id ) {
...
-			$svg    = @simplexml_load_file( $svg ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
-			$width  = 0;
-			$height = 0;
-			if ( $svg ) {
+			$svg      = get_attached_file( $attachment_id );
+			$metadata = wp_get_attachment_metadata( $attachment_id );
+			$width    = 0;
+			$height   = 0;
+
+			if ( $svg && ! empty( $metadata['width'] ) && empty( $metadata['height'] ) ) {
```
dkotter, line 594: @sksaju Sorry, I know this is late and this PR is already merged but in looking at this prior to prepping the next release, on the surface it seems like this conditional is not correct. It checks `$svg` isn't `null` and then checks if the width **isn't** empty and the height **is** empty. Unless I'm thinking of this wrong, shouldn't it be checking if both width and height aren't empty, so like this:

```php
if ( $svg && ! empty( $metadata['width'] ) && ! empty( $metadata['height'] ) ) {
```

sksaju: @dkotter Yeah it should be `! empty`, I've created a PR https://github.com/10up/safe-svg/pull/174 to fix it, Apologies & Thanks

PR #161, opened 2023-10-26
Update classname

tobeycodes:
### Description of the Change
Update the root class name to make `theme.json` changes possible. 

If you add this to your theme.json the css is applied to the block classname `wp-block-safe-svg-svg-icon` so this does not get applied to the block.

```
      "safe-svg/svg-icon": {
        "border": {
          "color": "var(--wp--preset--color--gray-200)",
          "radius": "8px",
          "style": "solid",
          "width": "1px"
        },
      }
```

### How to test the Change
Apply the theme.json change above against develop and this branch to see the difference

### Credits
@tobeycodes 

### Checklist:
- [X] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [X] I have updated the documentation accordingly.

### DIFF
--- a/includes/blocks/safe-svg/edit.js
+++ b/includes/blocks/safe-svg/edit.js
@@ -48,7 +48,7 @@ const SafeSvgBlockEdit = ( props ) => {
 	} = attributes;
 	const blockProps = useBlockProps(
 		{
-			className:` wp-block-safe-svg-svg-icon`,
+			className:` wp-block-safe-svg-svg-icon safe-svg-cover`,
 			style: {
 				textAlign: alignment,
 			}
--- a/includes/blocks/safe-svg/edit.js
+++ b/includes/blocks/safe-svg/edit.js
@@ -48,7 +48,7 @@ const SafeSvgBlockEdit = ( props ) => {
 	} = attributes;
 	const blockProps = useBlockProps(
 		{
-			className:` safe-svg-cover`,
+			className:` wp-block-safe-svg-svg-icon`,
 			style: {
 				textAlign: alignment,
 			}
--- a/includes/blocks/safe-svg/register.php
+++ b/includes/blocks/safe-svg/register.php
@@ -67,7 +67,7 @@ function render_block_callback( $attributes ) {
 	return apply_filters(
 		'safe_svg_inline_markup',
 		sprintf(
-			'<div class="safe-svg-cover" style="text-align: %s;">
+			'<div class="wp-block-safe-svg-svg-icon" style="text-align: %s;">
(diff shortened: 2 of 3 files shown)


includes/blocks/safe-svg/edit.js
```diff
@@ -48,7 +48,7 @@ const SafeSvgBlockEdit = ( props ) => {
 	} = attributes;
 	const blockProps = useBlockProps(
 		{
-			className:` safe-svg-cover`,
+			className:` wp-block-safe-svg-svg-icon`,
```
dkotter: Probably edge case but there may be developers that have added specific styling based on the `safe-svg-cover` class, so would probably be best to keep that class along with this new one I think
```suggestion
			className:` wp-block-safe-svg-svg-icon safe-svg-cover`,
```

includes/blocks/safe-svg/register.php
```diff
@@ -67,7 +67,7 @@ function render_block_callback( $attributes ) {
 	return apply_filters(
 		'safe_svg_inline_markup',
 		sprintf(
-			'<div class="safe-svg-cover" style="text-align: %s;">
+			'<div class="wp-block-safe-svg-svg-icon" style="text-align: %s;">
```
dkotter: Same suggestion as above
```suggestion
			'<div class="wp-block-safe-svg-svg-icon safe-svg-cover" style="text-align: %s;">
```

dkotter (changes requested):

includes/blocks/safe-svg/edit.js
tobeycodes: That's a valid point. I have updated this.

includes/blocks/safe-svg/register.php
tobeycodes: I have updated this.

dkotter (approved):

tobycodes:
> ### Description of the Change
> Update the root class name to make `theme.json` changes possible.
> 
> If you add this to your theme.json the css is applied to the block classname `wp-block-safe-svg-svg-icon` so this does not get applied to the block.
> 
> ```
>       "safe-svg/svg-icon": {
>         "border": {
>           "color": "var(--wp--preset--color--gray-200)",
>           "radius": "8px",
>           "style": "solid",
>           "width": "1px"
>         },
>       }
> ```
> 
> ### How to test the Change
> Apply the theme.json change above against develop and this branch to see the difference
> 
> ### Changelog Entry
> > Added support theme.json block changes
> 
> ### Credits
> @tobycodes
> 
> ### Checklist:
> * [x]  I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
> * [x]  I have updated the documentation accordingly.
> * [ ]  I have added tests to cover my change.
> * [ ]  All new and existing tests pass.

I think I was wrongly tagged in this :)

Great work though.

PR #173, opened 2024-02-05
ensure blocks are present before processing

### LINKED ISSUE #170: When svg optimization is turned on, a javascript error can occur that breaks saving content
### Describe the bug

As reported on [WordPress.org](https://wordpress.org/support/topic/pages-fails-to-update-when-optimizer-enabled/), if svg optimization is turned on (which is a feature added in #79), there's a scenario where saving content doesn't work.

In testing myself, I was able to reproduce and discover the problem. A javascript console error ends up being thrown (`n.blocks is not iterable`) which causes any other javascript to no longer execute (like the save action).

This error comes from this line: https://github.com/10up/safe-svg/blob/72a69f0f93e7f589f6113facb9db751a18626a8a/assets/js/admin/admin.js#L249

In looking at that code, it seems to listen for the save event and then it tries iterating through all the block content to find any svgs that need optimization. The problem here is if no block changes happened, this causes an error because we try iterating on an element that doesn't actually exist.

There may be a better fix for this but at the very least, we need to ensure the `changes.blocks` attribute exists before we try iterating over it.

### Steps to Reproduce

Follow the instructions shown in the video here: https://wordpress.org/support/topic/pages-fails-to-update-when-optimizer-enabled/

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

psorensen:
### Description of the Change

Closes #

### How to test the Change

1. Set `safe_svg_optimizer_enabled` filter to return true 
`add_filter( 'safe_svg_optimizer_enabled', '__return_true' );`
3. Create a new Page, populate the page title and featured image. Do not populate any blocks.
4. Publish the post.
5. Page should publish without normally without issue.

### Changelog Entry
Fixed fatal JS error when optimization is enabled and page is published without blocks.

### Credits

Props @tictag @psorensen

### Checklist:

### DIFF
--- a/assets/js/admin/admin.js
+++ b/assets/js/admin/admin.js
@@ -246,7 +246,7 @@ import {select, subscribe} from '@wordpress/data';
     subscribe(() => {
         if (editorStore.isSavingPost()) {
             const changes = editorStore.getPostEdits();
-            // maske sure we have blocks to process
+            // make sure we have blocks to process
             if (!Array.isArray(changes.blocks) || !changes.blocks.length) {
                 return;
             }
--- a/assets/js/admin/admin.js
+++ b/assets/js/admin/admin.js
@@ -246,6 +246,10 @@ import {select, subscribe} from '@wordpress/data';
     subscribe(() => {
         if (editorStore.isSavingPost()) {
             const changes = editorStore.getPostEdits();
+            // maske sure we have blocks to process
+            if (!Array.isArray(changes.blocks) || !changes.blocks.length) {
+                return;
+            }
             for (const changedBlock of changes.blocks) {
                 const blockName = changedBlock?.name ?? '';
                 const innerBlocks = changedBlock?.innerBlocks ?? [];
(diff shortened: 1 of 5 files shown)


assets/js/admin/admin.js
```diff
@@ -246,6 +246,10 @@ import {select, subscribe} from '@wordpress/data';
     subscribe(() => {
         if (editorStore.isSavingPost()) {
             const changes = editorStore.getPostEdits();
+            // maske sure we have blocks to process
```
dkotter:
```suggestion
            // make sure we have blocks to process
```

dkotter (approved):

coreyworrell:
Any estimate when this will be released? Having to constantly enable/disable the plugin or comment out the filter when editing content is becoming a hassle. Thanks very much.

psorensen:
@coreyworrell Hard to provide timing, but it might be worth checking out the 2.3.0 milestone https://github.com/10up/safe-svg/milestone/10 to see the list of open tasks the next release is dependent on. 

Additionally, you can use the `develop` branch of this repo to access the latest features/fixes.

dkotter:
@coreyworrell Hoping to get the next release finalized and out this week

PR #174, opened 2024-02-12
fix width and height metadata condition on svg_dimensions function

sksaju:
### Description of the Change

Remediation work on the PR https://github.com/10up/safe-svg/pull/154 

Closes #

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/tests/unit/test-safe-svg.php
+++ b/tests/unit/test-safe-svg.php
@@ -262,6 +262,53 @@ class SafeSvgTest extends TestCase {
 							),
 						),
 					),
+					array(
+						'file' => __DIR__ . '/files/svgNoDimensions.svg',
+						'sizes' => array(
+							'thumbnail' => array(
+								'width' => 150,
+								'height' => 150,
+								'crop' => 1,
+								'file' => 'svgNoDimensions.svg',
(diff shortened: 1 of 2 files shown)


tests/unit/test-safe-svg.php
```diff
@@ -262,6 +262,53 @@ public function test_one_pixel_fix() {
 							),
 						),
 					),
+					array(
```
peterwilsoncc, line 265: Is this what `wp_get_attachment_metadata()` is expected to return following your changes in this and the other PR? 

Using `svgNoDimensions.svg` I don't get a sizes array stored.

peterwilsoncc (commented):
The change looks good but I left a question inline about the tests and differences I am seeing locally.

tests/unit/test-safe-svg.php
iamdharmesh: @peterwilsoncc I have updated this as per the actual response from the WP for the `svgNoDimensions.svg` file.

iamdharmesh (approved):
LGTM.

PR #193, opened 2024-04-17
Add a filter around the return value from our `current_user_can_upload_svg` method

### LINKED ISSUE #192: Allow the `current_user_can_upload_svg` method to be filtered

In #76, we added settings allowing you to choose which user roles can upload svgs. When a role is selected, we add the `safe_svg_upload_svg` capability to those that have that role and check that when an svg is attempted to be uploaded.

There was a recent request around how to allow non-logged in users to upload svgs (for instance, if you have a front-end upload form). I don't know if that specific use case is something we need to support here but I do think it would be useful to add a new filter around the value returned by our `current_user_can_upload_svg`. This would allow others to change who can upload svgs, either making that more strict or less strict.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

This PR adds a new filter, `safe_svg_current_user_can_upload`, around the returned value from our `current_user_can_upload_svg` method. This filter allows others to have more fine-tune control on who can or cannot upload svgs. For instance, this would allow you to lock down uploads to specific users, instead of specific roles. Or would allow you to allow non-logged in users the ability to upload svgs.

Closes #192 

### How to test the Change

1. With this PR checked out, go to the Media Settings page and ensure your user account has permissions to upload
2. Test uploading an svg and ensure it works
3. Switch to an account that doesn't have permission to upload
4. Ensure svgs aren't allowed
5. Now use the new `current_user_can_upload_svg` and modify permissions, running through the tests again to ensure things work. As an example, could add `add_filter( 'safe_svg_current_user_can_upload', '__return_false' );` and ensure that no one can upload svgs, or `add_filter( 'safe_svg_current_user_can_upload', '__return_true' );` and ensure everyone can upload svgs.

### Changelog Entry

> Added - New filter, `safe_svg_current_user_can_upload`, allowing more control over who can upload svg files.

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -153,13 +153,24 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 		 */
 		public function current_user_can_upload_svg() {
 			$upload_roles = get_option( 'safe_svg_upload_roles', [] );
+			$can_upload   = false;
 
-			// Fallback to upload_files check for backwards compatibility.
 			if ( empty( $upload_roles ) ) {
-				return current_user_can( 'upload_files' );
+				// Fallback to upload_files check for backwards compatibility.
+				$can_upload = current_user_can( 'upload_files' );
+			} else {
+				// Use our custom capability if some upload roles are set.
+				$can_upload = current_user_can( 'safe_svg_upload_svg' );
 			}
 
-			return current_user_can( 'safe_svg_upload_svg' );
+			/**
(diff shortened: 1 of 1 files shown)


iamdharmesh (approved):
Thanks for working on this @dkotter. PR LGTM and it tests well.

shmaltz:
Thank you @dkotter !

PR #203, opened 2024-06-14
Fix/194 PHP Deprecated Warning

### LINKED ISSUE #194: PHP 8.3 Compatibility
### Describe the bug

There is a small PHP warning coming from your plugin when upgrading to PHP 8.3:

`Deprecated: Automatic conversion of false to array is deprecated in /wp-content/plugins/safe-svg/safe-svg.php on line 652`

### Steps to Reproduce

1. Update PHP to 8.3
2. Have the Safe SVG plugin installed & activated
3. Visit any page in the frontend
4. Check PHP logs 

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

This PR handles the case where the image_meta is not an array and sets it to an empty array to avoid PHP warnings.

Closes #194

### How to test the Change

As this is not a straightforward issue so it is difficult to write steps to regenerate the issue.

### Changelog Entry

> Fixed - Fixed - Handled PHP warning when the `$image_meta` is not an array.

### Credits
Props @drazenbebic, @kirtangajjar, @faisal-alvi , @dkotter 

### Checklist:

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -699,6 +699,10 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 		 */
 		public function disable_srcset( $image_meta, $size_array, $image_src, $attachment_id ) {
 			if ( $attachment_id && 'image/svg+xml' === get_post_mime_type( $attachment_id ) ) {
+				// Convert to array if not already to avoid PHP warnings.
+				if ( ! is_array( $image_meta ) ) {
+					$image_meta = array();
+				}
 				$image_meta['sizes'] = array();
 			}
 
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -698,11 +698,7 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 		 * @param int    $attachment_id The image attachment ID.
 		 */
 		public function disable_srcset( $image_meta, $size_array, $image_src, $attachment_id ) {
-			if ( $attachment_id && 'image/svg+xml' === get_post_mime_type( $attachment_id ) ) {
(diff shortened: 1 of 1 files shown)


safe-svg.php
```diff
@@ -699,6 +699,10 @@ protected function svg_dimensions( $attachment_id ) {
 		 */
 		public function disable_srcset( $image_meta, $size_array, $image_src, $attachment_id ) {
 			if ( $attachment_id && 'image/svg+xml' === get_post_mime_type( $attachment_id ) ) {
+				// Convert to array if not already to avoid PHP warnings.
+				if ( ! is_array( $image_meta ) ) {
+					$image_meta = array();
+				}
```
dkotter: I believe this will work but in thinking about this more, in a situation where somehow the `$image_meta` field isn't an array, I'm worried about the potential consequences to forcing that to be an array. In addition, in that scenario, seems like there's no need to do anything here, as the `sizes` key won't exist (since `$image_meta` isn't an array) and thus the `srcset` won't ever be created anyway.

I'd suggest instead the following change to the line above:

```php
if ( $attachment_id && 'image/svg+xml' === get_post_mime_type( $attachment_id ) && is_array( $image_meta ) ) {
```

dkotter (changes requested):

safe-svg.php
faisal-alvi: Agreed, it makes complete sense, thanks! (https://github.com/10up/safe-svg/pull/203/commits/7d4e730d83712655910feee3344acafe6cee8237)

dkotter (approved):

PR #216, opened 2024-07-22
Update how custom svg dimesions are applied

### LINKED ISSUE #6: SVG height and width attributes
Hi @darylldoyle 

What are your thoughts on trying to assign the default height and width attributes based on the image size used in the `wp_get_attachment_image()` function?

```
echo wp_get_attachment_image( $svg_id, [78, 78], true, [
     'class' => 'img-fluid',
     'alt' => 'icon',
] );
```
For example, when I  use the above in my template with your plugin activated my SVG is loaded with the height and width attributes set to the images default size. I, however, would like the height and width to be `78`x`78`. 

I ended up using the following filter in my theme to work around the issue... but i think this would be worth considering incorporating into the plugin. 
```
add_filter( 'wp_get_attachment_image_attributes', 'wwc_fix_direct_image_output', 11, 3 );
function wwc_fix_direct_image_output( $attr, $attachment, $size = 'thumbnail' ) {

	// If we're not getting a WP_Post object, bail early.
	// @see https://wordpress.org/support/topic/notice-trying-to-get-property-id/
	if ( ! $attachment instanceof WP_Post ) {
		return $attr;
	}

	$mime = get_post_mime_type( $attachment->ID );
	if ( 'image/svg+xml' === $mime ) {
		$default_height = 100;
		$default_width  = 100;

		$size = image_downsize($attachment->ID, $size);
		
		if (is_array($size) && sizeof($size) >= 3) {
			$attr['height'] = $size[2];
			$attr['width']  = $size[1];
		}
	}

	return $attr;
}```

### PR DESCRIPTION
**Pull request by gabriel-glo**
### Description of the Change

- Update how custom svg dimesions are applied for `get_image_tag` and `wp_get_attachment_image` functions 
- Separate the logic for defining custom dimensions in a `set_svg_dimension` function

Closes #6 

### How to test the Change

In any template output the image markup either by using `get_image_tag` or `wp_get_attachment_image` functions. 
Set the custom image size to either one of the registered image size names or a custom one via array `[$width, $height]`. 
Output should match the set dimensions.

### Changelog Entry

> Added - added set_svg_dimension function in order to reduce repetition
> Changed - changed how image dimensions were passed in `get_image_tag_override` and `one_pixel_fix` methods

### Credits

Props @gabriel-glo, @jeremymoore, @iamdharmesh, @dkotter

### Checklist:

### CODE REVIEW

File: safe-svg.php [hunk 1]
```diff
@@ -728,6 +717,37 @@ protected function str_ends_with( $haystack, $needle ) {
...
 		}
 
+		/**
+		 * Set custom width or height of the SVG image.
+		 *
+		 * @param int          $id   Image attachment ID.
+		 * @param string|array $size Size of image. Image size or array of width and height values
+		 *                                    (in that order). Default 'thumbnail'.
+		 *
+		 * @return int|false Width or height of the SVG image, or false if not found.
```
iamdharmesh:
```suggestion
		 * @return array|bool Width or height of the SVG image, or false if not found.
```

iamdharmesh (dismissed):
Thanks a lot for the PR @gabriel-glo. Code looks good and it tests well. (Added one minor doc update suggestion)

@dkotter The PR looks good to me. However, since we had 2-3 size/dimension-related PRs on this plugin earlier and you have better context on those, I wanted to have a quick review from your side to ensure these changes align with those PRs.

Thank you.

safe-svg.php
```diff
@@ -728,6 +717,37 @@ protected function str_ends_with( $haystack, $needle ) {
...
+		 */
+		protected function set_svg_dimensions( $id, $size ) {
+			$dimensions = $this->svg_dimensions( $id );
+
+			if ( is_array( $size ) ) {
+				$width  = $size[0];
+				$height = $size[1];
+			} elseif ( 'full' === $size && $dimensions ) {
```
dkotter: Because the output of `svg_dimensions` can be filtered, I think we'll need to be more defensive in a few places here to ensure 1 - `$dimensions` is an array and 2 - it contains the `height` and `width` keys

safe-svg.php
```diff
@@ -728,6 +717,37 @@ protected function str_ends_with( $haystack, $needle ) {
...
+		/**
+		 * Set custom width or height of the SVG image.
+		 *
+		 * @param int          $id   Image attachment ID.
+		 * @param string|array $size Size of image. Image size or array of width and height values
+		 *                                    (in that order). Default 'thumbnail'.
+		 *
+		 * @return int|false Width or height of the SVG image, or false if not found.
+		 */
+		protected function set_svg_dimensions( $id, $size ) {
```
dkotter: Would be nice to add unit tests to cover this new method

safe-svg.php
```diff
@@ -728,6 +717,37 @@ protected function str_ends_with( $haystack, $needle ) {
 			return 0 === substr_compare( $haystack, $needle, -$len, $len );
 		}
 
+		/**
+		 * Set custom width or height of the SVG image.
```
dkotter: Since this method doesn't actually set anything, it returns these values, we should probably update this

dkotter: And thinking about it more, might be worth renaming this method to be `get_svg_dimensions` instead, or something similar

dkotter (changes requested):

safe-svg.php
```diff
@@ -729,6 +718,37 @@ protected function str_ends_with( $haystack, $needle ) {
...
+		 *
+		 * @return array|bool Width or height of the SVG image, or false if not found.
+		 */
+		protected function get_svg_dimensions( $id, $size ) {
+			$dimensions = $this->svg_dimensions( $id );
+
+			if ( is_array( $size ) ) {
+				$width  = $size[0];
+				$height = $size[1];
+			} elseif   ( 'full' === $size && is_array( $dimensions ) && isset( $dimensions['width'], $dimensions['height'] ) ) {
```
dkotter:
```suggestion
			} elseif ( 'full' === $size && is_array( $dimensions ) && isset( $dimensions['width'], $dimensions['height'] ) ) {
```

dkotter (changes requested):

dkotter (approved):
This looks good to me now. Tested with both `get_image_tag` and `wp_get_attachment_image` and verified things worked as expected with both. Testing on the latest release and on this PR, `get_image_tag` returns the exact same output. Testing `wp_get_attachment_image`, on the latest release, if you pass in an array of sizes, those values aren't being used. On this PR, those sizes are now respected and set properly as the height and width attributes.

martinpl:
This breaks behavior in wp_get_attachment_image(): 
before: any thumbnail size will inherit SVG width, height attributes
now: only 'full' thumbnail will inherit SVG width, height attributes

Reverting this helps:
https://github.com/10up/safe-svg/pull/216/files#diff-05bf19b873cfb6358cd7c239366db6675635bf4b03a4b56353102e359d5c2859L447

PR #228, opened 2024-09-26
Fix/only allow svg when uploading

darylldoyle:
### Description of the Change
Updates the plugin to only add SVGs as an allowed type, when the sanitiser will be able to run on those files.

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability
> Developer - Non-functional update

### Credits

### Checklist:

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -129,6 +129,7 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 			$this->sanitizer->minify( true );
 
 			add_action( 'init', array( $this, 'setup_blocks' ) );
+			add_action( 'load-upload.php', array( $this, 'allow_svg_from_upload' ) );
 			add_filter( 'wp_handle_sideload_prefilter', array( $this, 'check_for_svg' ) );
 			add_filter( 'wp_handle_upload_prefilter', array( $this, 'check_for_svg' ) );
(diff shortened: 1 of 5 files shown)


dkotter (changes requested):
So this doesn't work for me, I always get the `Sorry, you are not allowed to upload this file type` error message.

I think what's happening is there's an earlier mime check that fires before the `wp_handle_upload_prefilter` filter and since we aren't adding the svg mime until that filter runs, things no longer work.

To test, I'm in the Media Library admin page, with the Grid View on. If I drag and drop an svg image, I get the above error.

dkotter:
I was curious on how our E2E tests were passing here when I couldn't get things to work. Looking into that, the `uploadMedia` command we use goes to `/wp-admin/media-new.php` and uploads from there. If I go to that page directly, I can upload svgs just fine but as mentioned above, I can't upload svgs from the main Media Library screen, which is interesting

darylldoyle:
Thanks for flagging @dkotter! I too could replicate on the media grid view. I've resolved this and added additional tests for that. I also had to allow uploads from a few additional contexts, because otherwise you could _only_ upload SVGs from the media pages. We _might_ want to add some additional tests for these, but I don't have time at the moment.

dkotter (approved):

PR #238, opened 2024-12-03
Revert changes done in #216

### LINKED ISSUE #237: Breaking behavior in inheriting width, height attributes from 2.3.0
### Describe the bug

Hello,
Here are details:
https://github.com/10up/safe-svg/pull/216#issuecomment-2503662519

Thanks :)

### Steps to Reproduce

wp_get_attachment_image() without full thumbnail will not get original width and height

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

As reported in #237, the changes made in #216 are causing issues for anyone using the `wp_get_attachment_image` function to output SVGs. As described in that Issue, we need to be smarter in how we determine the height and width attributes to use instead of just relying on the dimensions for each registered size (since SVGs don't get cropped the same as normal images by WordPress).

This PR reverts the changes introduced in #216 and also adds some additional E2E tests to help us catch these sorts of issues in the future.

Note there is additional work we could look to do here, both fixing what we were attempting to do in #216 and being smarter with how we determine the SVG sizes, as described in this [comment](https://github.com/10up/safe-svg/issues/237#issuecomment-2514800549).

Closes #237

### How to test the Change

1. Ensure general functionality of the plugin still works as expected (you can upload svgs, you can add those into content, etc)
2. Also modify your theme to directly output svgs using `wp_get_attachment_image` and ensure that no matter what image size you pass in, the full SVG dimensions are used instead (which is previous behavior)

### Changelog Entry

> Fixed - Revert changes made to how we determine custom dimensions for svgs

### Credits

Props @dkotter, @martinpl, @subfighter3, @smerriman, @gigatyrant

### Checklist:

### DIFF
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -504,8 +504,7 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 				if ( is_array( $size ) ) {
 					$width  = $size[0];
 					$height = $size[1];
-				// phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.Found, Squiz.PHP.DisallowMultipleAssignments.FoundInControlStructure
-				} elseif ( 'full' === $size && $dimensions = $this->svg_dimensions( $id ) ) {
+				} elseif ( 'full' === $size && $dimensions = $this->svg_dimensions( $id ) ) { // phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.Found, Squiz.PHP.DisallowMultipleAssignments.FoundInControlStructure
 					$width  = $dimensions['width'];
 					$height = $dimensions['height'];
 				} else {
--- a/safe-svg.php
+++ b/safe-svg.php
@@ -256,7 +256,7 @@ if ( ! class_exists( 'SafeSvg\\safe_svg' ) ) {
 				return $file;
 			}
 
-			$file_name   = isset( $file['name'] ) ? $file['name'] : '';
+			$file_name = isset( $file['name'] ) ? $file['name'] : '';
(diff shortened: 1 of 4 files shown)


jeffpaul (dismissed):

dkotter:
This reverts the changes made in #216 and brings back the functionality for `wp_get_attachment_image`. In testing with `get_image_tag` though, that doesn't seem to be working as expected anymore, though seems that's a WordPress core issue and not something we changed (I tested Safe SVG back to 2.0.0 and got the same results).

Basically that function is using the dimensions of the image size passed in instead of the SVG dimensions (basically functions the way `wp_get_attachment_image` did after the changes in #216, which we're reverting here). Don't need to hold up this PR for that but we should open a new issue to investigate what needs to change to ensure `get_image_tag` outputs SVGs as expected. Seems the problem lies in [our code](https://github.com/10up/safe-svg/blob/1df2fedf77db13029584ea7ed104a86fa29190f9/safe-svg.php#L505-L511) that looks for height and width set to 1 and replaces those with the custom size. WordPress isn't setting those to 1 anymore so that replacement doesn't work.

I've added E2E tests in this PR for both `wp_get_attachment_image` and `get_image_tag` but the `get_image_tag` ones, while they pass, aren't getting the height and width attributes I'd expect, so once that issue gets fixed, those tests should fail (which is good in this case) and we can fix the tests at that point.

jeffpaul (approved):

iamdharmesh (approved):
Code changes look good to me and it tests well. Thanks @dkotter

PR #261, opened 2025-06-28
Add icon fill color support

s3rgiosan:
### Description of the Change
This PR adds a new "Icon color" setting, allowing editors to control the SVG fill color. It also removes text color support, as it's redundant in this icon-specific context.

Before

Screenshot 2025-06-28 at 17 31 02

After

Screenshot 2025-06-28 at 17 31 30

### Changelog Entry

> Added - fillColor support with Icon color settings

### Credits

Props @s3rgiosan

### Checklist:

- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.

### DIFF
--- a/includes/blocks/safe-svg/block.json
+++ b/includes/blocks/safe-svg/block.json
@@ -47,7 +47,7 @@
   "supports": {
     "html": false,
     "color": {
-      "text": true,
+      "text": false,
       "background": true
     },
     "spacing": {
--- a/includes/blocks/safe-svg/block.json
+++ b/includes/blocks/safe-svg/block.json
@@ -36,6 +36,12 @@
     },
     "imageSizes": {
       "type": "object"
+    },
+    "iconColor": {
+      "type": "string"
+    },
+    "iconColorValue": {
+      "type": "string"
     }
   },
   "supports": {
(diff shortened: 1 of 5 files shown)


dkotter (approved):

dkotter:
@s3rgiosan Apologies for the ping here after this is already merged but in doing final testing today in advance of doing a release (#263) I noticed that if I set a custom icon color, that shows properly in the editor but doesn't show on the front-end (which I must have missed when testing this previously).

It seems on the front-end (in `register.php`) we still use the `textColor` attribute, which no longer exists and needs updated to I believe `iconColor`. In addition, wondering if we need to consider supporting sites that did previously have `textColor` set, instead of just replacing that with `iconColor`. Basically wondering if we need a fallback to look at `textColor` first and use it if it exists and if not, fallback to `iconColor`?

Wondering if you have time to open a new PR to address this to help unblock the release?

s3rgiosan:
@dkotter I'll work on a fix as soon as possible

s3rgiosan:
@dkotter This was addressed [here](https://github.com/10up/safe-svg/pull/266). 

I reverted the change that added the icon color. While using the icon color makes more sense than relying on the text color for the SVG, the goal here was to unblock this release. We can revisit and properly add icon color support later, including fallback to text color.

Let me know if you'd prefer to hold off and address the icon color implementation before releasing the new version.

PR #272, opened 2025-08-12
Bump svg sanitization library

darylldoyle:
### Description of the Change

Updated the `enshrined/svg-sanitize` package from version `^0.21.0` to `^0.22.0` in `composer.json` to use the latest release.

This updated resolves an issue where attribute name case inconsistencies (e.g., `xlink:Href` instead of `xlink:href`) in XML processing could cause namespace lookups and sanitisation to fail. Attribute names in both namespaced and non-namespaced contexts are now normalised to their expected lowercase form before processing. This ensures consistent sanitisation of `xlink:href` and other targeted attributes regardless of their original case.

### Changelog Entry

* Security - Updated the sanitisation library to fix an issue with case-insensitive attributes slipping through the sanitiser.

### Credits
@darylldoyle 

### Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
 	],
 	"require": {
 		"php": ">=7.4",
-		"enshrined/svg-sanitize": "^0.21.0"
+		"enshrined/svg-sanitize": "^0.22.0"
 	},
 	"require-dev": {
 		"10up/phpcs-composer": "dev-master",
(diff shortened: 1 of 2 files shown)


dkotter (approved):
Tested with an svg that had an anchor tag with the attribute `xlink:hReF`. Before this PR, that attribute was not being sanitized. After pulling this PR and running `composer install`, the attribute does get sanitized properly

PR #273, opened 2025-08-12
Fix/argument count error get image tag override

markmclaren:
### Description of the Change

This PR fixes a fatal `ArgumentCountError` that occurs when the `get_image_tag_override` method is called with fewer than the expected 6 arguments:

```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function SafeSvg\safe_svg::get_image_tag_override(), 2 passed in /var/www/html/wp-includes/class-wp-hook.php on line 324 and exactly 6 expected
```

**Root Cause Analysis:**
1. **Incorrect Hook Type**: The `get_image_tag` was registered as an action instead of a filter
2. **Rigid Parameter Requirements**: The method required exactly 6 parameters with no defaults
3. **No Fallback Handling**: There was no graceful handling when fewer arguments were provided

**Solution:**
1. **Corrected Hook Registration**: Changed from `add_action` to `add_filter` for the `get_image_tag` hook (which is a filter according to [WordPress documentation](https://developer.wordpress.org/reference/hooks/get_image_tag/))
2. **Made Parameters Optional**: Added default values to all parameters except the first (`$html`)
3. **Added Early Return Safety Check**: Returns HTML unchanged if no valid attachment ID is provided

**Benefits:**
- Prevents fatal errors in edge cases or different WordPress environments
- Maintains backward compatibility with existing functionality
- Improves robustness against plugin conflicts or WordPress core changes
- Preserves accessibility features (adds `role="img"` for SVGs)

**Verification Steps:**
- All existing unit tests continue to pass
- Added comprehensive test coverage for edge cases
- Method now handles both standard 6-parameter calls and edge cases with fewer arguments

Closes #

### How to test the Change

1. **Run the existing test suite**: `composer test:unit` - all tests should pass
2. **Test the specific edge case**: Run the new test cases that verify the method handles fewer arguments gracefully
3. **Test normal SVG functionality**: Upload an SVG and verify it displays correctly with proper dimensions and accessibility attributes
4. **Test the fix in a WordPress environment**: 
   - Install the plugin in a WordPress site
   - Upload SVG files and verify no fatal errors occur
   - Check that SVGs display correctly in media library and frontend

### Changelog Entry

Fixed - ArgumentCountError in get_image_tag_override method when called with fewer than 6 arguments

### Credits

Props @markmclaren

### Checklist:


markmclaren:
@jeffpaul I see this is assigned to me.  It passes the tests and I have tested it on WordPress.  Is any further action needed?

safe-svg.php
```diff
@@ -142,7 +142,7 @@ public function __construct() {
 			add_filter( 'wp_get_attachment_image_src', array( $this, 'one_pixel_fix' ), 10, 4 );
 			add_filter( 'admin_post_thumbnail_html', array( $this, 'featured_image_fix' ), 10, 3 );
 			add_action( 'admin_enqueue_scripts', array( $this, 'load_custom_admin_style' ) );
-			add_action( 'get_image_tag', array( $this, 'get_image_tag_override' ), 10, 6 );
+			add_filter( 'get_image_tag', array( $this, 'get_image_tag_override' ), 10, 6 );
```
dkotter, line 145: Nice catch! Technically behind the scenes, WordPress converts all `add_action` calls to `add_filter` calls so this shouldn't cause any issues but still nice to have this be the correct hook type

dkotter (commented):
Changes here look good to me but wondering if you have more details on how to reproduce the problem? The filter WordPress runs on `get_image_tag` will always return 6 arguments, so seems like in normal circumstances, this wouldn't ever be a problem. My guess is some other bit of code (maybe another plugin) is calling `get_image_tag` and not passing in valid values? Definitely good to be a bit more defensive on our side to account for that, just wondering if there are more specifics so I can reproduce and test this myself?

markmclaren:
@dkotter Ah, that makes perfect sense! I just discovered the root cause - I have a local custom taxonomy called image_tag which is creating a naming conflict. The issue isn't with WordPress's core get_image_tag filter, but rather this custom taxonomy interfering with the expected behavior.

Sorry for not catching that earlier - sometimes the most obvious causes are the hardest to spot.

PR #276, opened 2025-08-13
Fix a typo in dependency-review.yml

szepeviktor:


.github/workflows/dependency-review.yml
```diff
@@ -1,6 +1,6 @@
 # Dependency Review Action
 #
-# This Action will scan dependency manifest files that change as part of a Pull Reqest, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
+# This Action will scan dependency manifest files that change as part of a Pull Request, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
```
jeffpaul, line 3: Entirely likely that this is the same in all other 10up repos using this action.  Sigh.

jeffpaul (approved):

.github/workflows/dependency-review.yml
szepeviktor: Just introduce https://github.com/crate-ci/typos in 10up.

PR #315, opened 2026-05-23
Add link support to SVG Inline block

### LINKED ISSUE #282: SVG Inline Block, support link

I'd like to use SVG inline Block and like to link it exactly the way you can link a image.
If there's interest to have that feature, I could create a PR.

### Designs

Image

_No response_

### Code of Conduct

mgiannopoulos24:
### Description of the Change
Added `href`, `linkTarget`, `nofollow`, and `sponsored` attributes to the Safe SVG block, allowing users to wrap inline SVGs in links — similar to the core Image block.

Closes #282 

### How to test the Change
1. Insert an Inline SVG block and select an SVG
2. Click the link button (chain icon) in the block toolbar
3. Enter a URL, toggle Open in new tab, Add rel="nofollow", and/or Add rel="sponsored"
4. Publish the post and verify the SVG is wrapped in an <a> tag with correct href, target, and rel attributes

### Credits
Props @vegetable-bits

### Checklist:
- [X] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [X] I have updated the documentation accordingly.
- [X] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.
- [X] All new and existing tests pass.

### DIFF
--- a/includes/blocks/safe-svg/block.json
+++ b/includes/blocks/safe-svg/block.json
@@ -42,7 +42,7 @@
     },
     "linkTarget": {
       "type": "string",
-      "default": "_self"
+      "default": ""
     },
     "nofollow": {
       "type": "boolean",
--- a/includes/blocks/safe-svg/block.json
+++ b/includes/blocks/safe-svg/block.json
@@ -51,6 +51,9 @@
     "sponsored": {
       "type": "boolean",
       "default": false
+    },
+    "linkLabel": {
+      "type": "string"
     }
   },
   "supports": {
@@ -65,5 +68,6 @@
     }
   },
   "editorScript": "file:../../../dist/safe-svg-block.js",
+  "editorStyle": "file:../../../dist/safe-svg-block.css",
   "style": "file:../../../dist/safe-svg-block-frontend.css"
 }
--- a/includes/blocks/safe-svg/block.json
+++ b/includes/blocks/safe-svg/block.json
@@ -36,6 +36,21 @@
     },
     "imageSizes": {
       "type": "object"
+    },
+    "href": {
+      "type": "string"
+    },
+    "linkTarget": {
+      "type": "string",
+      "default": "_self"
+    },
+    "nofollow": {
+      "type": "boolean",
+      "default": false
+    },
+    "sponsored": {
+      "type": "boolean",
+      "default": false
     }
   },
   "supports": {
--- a/includes/blocks/safe-svg/edit.js
+++ b/includes/blocks/safe-svg/edit.js
@@ -15,6 +15,7 @@ import {
 	PanelBody,
 	Dropdown,
 	ToolbarButton,
+	TextControl,
 } from '@wordpress/components';
 import { link } from '@wordpress/icons';
 import {
@@ -59,6 +60,7 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
 		linkTarget,
 		nofollow,
 		sponsored,
+		linkLabel,
 	} = attributes;
 
 	const blockProps = useBlockProps(
@@ -251,6 +253,12 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
 										]}
 										onClose={onClose}
 									/>
+									<TextControl
+										label={__('Link Label (aria-label)', 'safe-svg')}
+										value={linkLabel}
+										onChange={(value) => setAttributes({ linkLabel: value })}
+										help={__('Provides an accessible label for screen readers.', 'safe-svg')}
+									/>
 								</div>
 							)}
 						/>
@@ -274,23 +282,24 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
 			{svgURL &&
 				<div {...containerBlockProps}>
 					{href ? (
-						<a
-							href={href}
-							target={linkTarget}
-							rel={[nofollow ? 'nofollow' : '', sponsored ? 'sponsored' : ''].filter(Boolean).join(' ') || undefined}
+						<div
+							style={style}
+							className={classnames(
+								'safe-svg-inside',
+								getColorClassName('color', textColor) || ''
+							)}
 						>
-							<div
-								style={style}
-								className={classnames(
-									'safe-svg-inside',
-									getColorClassName('color', textColor) || ''
-								)}
+							<a
+								href={href}
+								target={linkTarget}
+								rel={[nofollow ? 'nofollow' : '', sponsored ? 'sponsored' : '', linkTarget === '_blank' ? 'noopener' : ''].filter(Boolean).join(' ') || undefined}
+								aria-label={linkLabel || undefined}
 							>
 								<ReactSVG src={svgURL} beforeInjection={(svg) => {
 									svg.setAttribute('style', `width: ${dimensionWidth}px; height: ${dimensionHeight}px;`);
 								}} />
-							</div>
-						</a>
+							</a>
+						</div>
 					) : (
 						<div
 							style={style}
@@ -337,6 +346,7 @@ SafeSvgBlockEdit.propTypes = {
 		linkTarget: PropTypes.string,
 		nofollow: PropTypes.bool,
 		sponsored: PropTypes.bool,
+		linkLabel: PropTypes.string,
 	}).isRequired,
 	className: PropTypes.string,
 	clientId: PropTypes.string,
--- a/includes/blocks/safe-svg/edit.js
+++ b/includes/blocks/safe-svg/edit.js
@@ -13,7 +13,10 @@ import { __ } from '@wordpress/i18n';
 import {
 	Placeholder,
 	PanelBody,
+	Dropdown,
+	ToolbarButton,
 } from '@wordpress/components';
+import { link } from '@wordpress/icons';
 import {
 	useBlockProps,
 	BlockControls,
@@ -22,6 +25,7 @@ import {
 	__experimentalImageSizeControl as ImageSizeControl,
 	MediaReplaceFlow,
 	MediaPlaceholder,
+	LinkControl,
 	getColorClassName,
 } from '@wordpress/block-editor';
 
@@ -51,6 +55,10 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
 		dimensionWidth,
 		dimensionHeight,
 		textColor,
+		href,
+		linkTarget,
+		nofollow,
+		sponsored,
 	} = attributes;
 
 	const blockProps = useBlockProps(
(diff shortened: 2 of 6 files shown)


includes/blocks/safe-svg/edit.js
```diff
@@ -210,17 +273,37 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
-							'safe-svg-inside',
-							getColorClassName('color', textColor) || ''
-						)}
-					>
-						<ReactSVG src={svgURL} beforeInjection={(svg) => {
-							svg.setAttribute('style', `width: ${dimensionWidth}px; height: ${dimensionHeight}px;`);
-						}} />
-					</div>
+					{href ? (
+						<a
```
thrijith: `<a>` wraps a non-text SVG with no accessible name, it will fail WCAG 2.1 A. Should we add a `linkLabel` attr (TextControl in the link UI) rendered as `aria-label`?

Same for https://github.com/10up/safe-svg/pull/315/changes#diff-bb2bf39c5dff7589242bca4b6837a0a3ac99c025249a8c9ac81c524c1abcce3eR62-R68

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,22 @@ function render_block_callback( $attributes ) {
...
 
+	if ( ! empty( $attributes['href'] ) ) {
+		$link_target = ! empty( $attributes['linkTarget'] ) ? $attributes['linkTarget'] : '_self';
+		$rel         = trim(
+			( ! empty( $attributes['nofollow'] ) ? 'nofollow' : '' )
+			. ' '
+			. ( ! empty( $attributes['sponsored'] ) ? 'sponsored' : '' )
+		);
+		$contents    = sprintf(
+			'<a href="%1$s" target="%2$s" rel="%3$s">%4$s</a>',
```
thrijith: The `<a>` is added at different depths per layer: 

`edit.js` → `a > .safe-svg-inside > svg`.
`register.php` → `.safe-svg-inside > a > svg`. 

So `.wp-block-safe-svg-svg-icon a { align-self:center }` works only in the editor, not the front end, editor preview won't match output. 

Should `edit.js` wrap only the SVG inside `.safe-svg-inside`, mirroring the PHP nesting?

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,22 @@ function render_block_callback( $attributes ) {
 	 */
 	$class_name = apply_filters( 'safe_svg_inline_class', 'safe-svg-inline' );
 
+	if ( ! empty( $attributes['href'] ) ) {
+		$link_target = ! empty( $attributes['linkTarget'] ) ? $attributes['linkTarget'] : '_self';
+		$rel         = trim(
```
thrijith: `rel` is built from `nofollow`/`sponsored` only. 

Should we add `noopener` when `linkTarget==='_blank'` to match core and harden old browsers?

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,22 @@ function render_block_callback( $attributes ) {
...
+			( ! empty( $attributes['nofollow'] ) ? 'nofollow' : '' )
+			. ' '
+			. ( ! empty( $attributes['sponsored'] ) ? 'sponsored' : '' )
```
thrijith: It will output `rel` when empty; editor omits it.

Image


Should build the attr conditionally `$rel ? ' rel="'.esc_attr($rel).'"' : ''` ?

includes/blocks/safe-svg/frontend.scss
```diff
@@ -1,3 +1,17 @@
+.wp-block-safe-svg-svg-icon a {
+    align-self: center;
+}
+
+.components-dropdown {
```
thrijith: `block.json` maps `frontend.scss` to `style` (loads in editor + front end no `editorStyle`). 

The un-scoped `.components-dropdown { align-self:center }` hits unrelated editor dropdown, and the toolbar-only rules also ship to the front end as dead CSS. 

Should it scope to `.safe-svg-link-dropdown` and move to an editor-only sheet (`editor.scss` + `"editorStyle"`)?

thrijith (changes requested):

jeffpaul:
Just approved workflows to run, @thrijith if you can re-code review as it looks like updates here may resolve what you've called out so far

thrijith (dismissed):
Thanks a lot for the quick turn around, great work @mgiannopoulos24 !

jeffpaul:
@dkotter @peterwilsoncc either of you want a final review or good to merge from here?

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,34 @@ function render_block_callback( $attributes ) {
...
+
+		$rel_parts = array();
+		if ( ! empty( $attributes['nofollow'] ) ) {
+			$rel_parts[] = 'nofollow';
+		}
+		if ( ! empty( $attributes['sponsored'] ) ) {
+			$rel_parts[] = 'sponsored';
+		}
+		if ( '_blank' === $link_target ) {
+			$rel_parts[] = 'noopener';
```
dkotter: I think we should set this to `noreferrer noopener` or is there a specific reason to not have `noreferrer` here?

includes/blocks/safe-svg/edit.js
```diff
@@ -191,6 +201,67 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
+								<ToolbarButton
+									icon={link}
+									label={__('Link', 'safe-svg')}
+									onClick={onToggle}
+									aria-expanded={isOpen}
+								/>
+							)}
+							renderContent={({ onClose }) => (
+								<div className="block-editor-link-control">
+									<LinkControl
```
dkotter, line 216: I don't see a way to actually clear out a link after it's added, unless I'm missing it?

includes/blocks/safe-svg/edit.js
```diff
@@ -210,17 +281,38 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
-					</div>
+					{href ? (
+						<div
+							style={style}
+							className={classnames(
+								'safe-svg-inside',
+								getColorClassName('color', textColor) || ''
+							)}
+						>
+							<a
```
dkotter: Why do we need to render the anchor tag in the editor? The issue now is I can't click on the SVG Icon block because instead I'm clicking on the link itself. I would suggest we don't need to render this in the editor, only the frontend

dkotter (changes requested):

includes/blocks/safe-svg/register.php
peterwilsoncc: @dkotter @mgiannopoulos24 I think this can be removed alltogether for the front end. In recent years, browsers made `noopener` the default for links opening in new tabs, see the [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener).

For the front end I think it's polite to send along the referrer. In the `Edit` block, I'd always add `noreferrer` as some sites may not wish their admin pages to be shared with third parties.

includes/blocks/safe-svg/edit.js
```diff
@@ -191,6 +201,67 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
+										}}
+										onChange={(linkSettings) => {
+											setAttributes({
+												href: linkSettings.url,
+												linkTarget: linkSettings.opensInNewTab ? '_blank' : '_self',
+												nofollow: !!linkSettings.nofollow,
+												sponsored: !!linkSettings.sponsored,
+											});
+										}}
+										settings={[
```
peterwilsoncc, line 239: Adds a remove link, as noted by Darin
```suggestion
										onRemove={() => {
											setAttributes({
												href: '',
												linkTarget: '_self',
												nofollow: false,
												sponsored: false,
											});
										}}
										settings={[
```

includes/blocks/safe-svg/block.json
```diff
@@ -36,6 +36,24 @@
     },
     "imageSizes": {
       "type": "object"
+    },
+    "href": {
+      "type": "string"
+    },
+    "linkTarget": {
+      "type": "string",
+      "default": "_self"
```
peterwilsoncc:
```suggestion
      "default": ""
```

includes/blocks/safe-svg/edit.js
```diff
@@ -191,6 +201,75 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
+										value={{
+											url: href,
+											opensInNewTab: linkTarget === '_blank',
+											nofollow: !!nofollow,
+											sponsored: !!sponsored,
+										}}
+										onChange={(linkSettings) => {
+											setAttributes({
+												href: linkSettings.url,
+												linkTarget: linkSettings.opensInNewTab ? '_blank' : '_self',
```
peterwilsoncc:
```suggestion
												linkTarget: linkSettings.opensInNewTab ? '_blank' : '',
```

includes/blocks/safe-svg/edit.js
```diff
@@ -191,6 +201,75 @@ const SafeSvgBlockEdit = ({ attributes, setAttributes }) => {
...
+												nofollow: !!linkSettings.nofollow,
+												sponsored: !!linkSettings.sponsored,
+											});
+										}}
+										onRemove={() => {
+											setAttributes({
+												href: '',
+												linkTarget: '_self',
```
peterwilsoncc:
```suggestion
												linkTarget: '',
```

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,31 @@ function render_block_callback( $attributes ) {
 	 */
 	$class_name = apply_filters( 'safe_svg_inline_class', 'safe-svg-inline' );
 
+	if ( ! empty( $attributes['href'] ) ) {
+		$link_target = ! empty( $attributes['linkTarget'] ) ? $attributes['linkTarget'] : '_self';
```
peterwilsoncc:
```suggestion
		$link_target = ! empty( $attributes['linkTarget'] ) ? $attributes['linkTarget'] : false;
```

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,31 @@ function render_block_callback( $attributes ) {
...
+		$aria_label = ! empty( $attributes['linkLabel'] ) ? ' aria-label="' . esc_attr( $attributes['linkLabel'] ) . '"' : '';
+
+		$contents = sprintf(
+			'<a href="%1$s" target="%2$s"%3$s%4$s>%5$s</a>',
+			esc_url( $attributes['href'] ),
+			esc_attr( $link_target ),
+			$rel_attr,
+			$aria_label,
+			$contents
+		);
```
peterwilsoncc, line 77:
```suggestion
		$contents = sprintf(
			'<a href="%1$s"%2$s%3$s%4$s>%5$s</a>',
			esc_url( $attributes['href'] ),
			$link_target ? ' target="' . esc_attr( $link_target ) . '"' : '',
			$rel_attr,
			$aria_label,
			$contents
		);
```

peterwilsoncc (changes requested):
I've added one final suggestion inline over multiple parts.

The change is to remove the target instead of using `target="_self"` when the link isn't set to open in a new window.

Apart from that I think this is good to go.

peterwilsoncc (approved):
This is testing well and the code looks good to me.

The E2E test failures in trunk are unrelated to this change so I will merge as is.

dkotter (approved):

pbiron:
yet another feature request I was coming here to make, but I see you guys are already on top of it. It's scary how well y'all read my mind ;-)

PR #327, opened 2026-09-01
Update SVG Sanitizer library from v0.22.0 to v1.0.0

dkotter:
### Description of the Change

Updated the `enshrined/svg-sanitize` package from version `^0.22.0` to `^1.0.0` in composer.json to ensure we use the latest release.

There's a number of fixes in that [release](https://github.com/darylldoyle/svg-sanitizer/releases/tag/1.0.0) so it will be good to pull those in.

### How to test the Change

This update pulls in a new version of the library we use for sanitization so basic smoke testing of the entire plugin is needed, ensuring SVG uploads still work as expected.

### Changelog Entry

> Security - Updates the underlying `enshrined/svg-sanitize` package to pull in a security fixes

### Credits

Props @dkotter

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.
- [x] All new and existing tests pass.

PR #328, opened 2026-09-01
Render the SVG block in a shadow DOM

dkotter:
### Description of the Change

When rendering an SVG in our SVG block, we don't strip out any styles. This can result in someone adding styles that impact other areas of the page and by rendering the SVG, they can impact how the rest of the page looks.

To address this, this PR introduces new handling within the SVG block for any SVG that comes with it's own styles, wrapping those SVGs in a shadow DOM which keeps the styles isolated. This only applies to SVGs that come with styles (though can be turned on/off using the new `safe_svg_inline_use_shadow_dom` filter) so for majority of use cases, this won't change anything.

Note though for SVGs that do ship with their own styles, theme CSS will no longer be able to target them because of this shadow DOM. That styling will need to be done from within the SVG itself or you can opt out using the `safe_svg_inline_use_shadow_dom` filter.

In addition, replace the use of `ReactSVG` with our own custom `InlineSvg` component that will offer the same protections within the block editor.

We've also introduced the new `removeRemoteReferences` property of the sanitizer (coming in v1.0.0) and set that to false by default. This can be turned on with the new `safe_svg_remove_remote_references` filter, which will then strip out any remote references, making the SVG markup even more secure.

### How to test the Change

**NOTE**: The code here is dependent on the update in https://github.com/10up/safe-svg/pull/327 so that will either need merged first before testing this or you'll need to manually pull that change in locally when testing.

Basic testing across the plugin is desired here to ensure these changes don't break any existing functionality. Main thing to test is the SVG block.

Beyond that, the automated tests this PR ships with contains some SVG files you can use for testing to ensure the styles they load don't leak into the main document.

### Changelog Entry

> Added - New `safe_svg_remove_remote_references` filter to strip remote `url()`, `@import` and `image-set()` references, along with remote `href` targets, from uploaded SVGs. Off by default, because legitimate SVGs reference remote fonts and images but use this filter to turn it on

> Added - New `safe_svg_inline_use_shadow_dom` filter to control which inline SVGs are isolated in a shadow root, and new `safe_svg_inline_shadow_styles` filter to adjust the CSS injected alongside them

> Changed - Theme CSS can no longer target an inline SVG that carries its own `<style>` element, because stylesheets cannot reach into a shadow root. Style those SVGs from within the SVG itself, or opt out with the `safe_svg_inline_use_shadow_dom` filter. Inherited properties, including `color`/`currentColor` and custom properties, still apply as before, and SVGs without a `<style>` element are unaffected

> Security - The Inline SVG block now renders SVGs that carry their own `<style>` element inside a shadow root, so their CSS is scoped to the block instead of applying to the whole page

> Developer - Replaced the `react-svg` dependency with a small internal component, so the block editor preview isolates SVG CSS the same way the front end does

### Credits

Props @darylldoyle, @dkotter 

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,18 @@ add_filter( 'svg_allowed_tags', function ( $tags ) {
 } );
 ```
 
+### Can my theme style an inline SVG?
+
+Mostly, yes. The Inline SVG block renders an SVG that carries its own `<style>` element inside a shadow root, because CSS inside an inline SVG is otherwise applied to the whole page rather than just the SVG. Stylesheets cannot reach into a shadow root, so theme CSS such as `.entry-content svg { fill: red; }` will not apply to those SVGs.
+
+Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered exactly as before and can be styled normally.
+
+To turn isolation off, at the cost of allowing an SVG's CSS to affect the rest of the page:
+
+```php
+add_filter( 'safe_svg_inline_use_shadow_dom', '__return_false' );
+```
+
 ### Why doesn't Safe SVG globally enable SVG uploads?
 
 Safe SVG only allows SVGs through upload paths it can actively sanitize. While most WordPress uploads use standard functions like `wp_handle_upload()` (which Safe SVG hooks), plugins and themes can create custom upload paths by calling WordPress's underlying `_wp_handle_upload()` function with arbitrary action parameters.
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ add_filter( 'svg_allowed_tags', function ( $tags ) {
 
 Mostly, yes. The Inline SVG block renders an SVG that carries its own `<style>` element inside a shadow root, because CSS inside an inline SVG is otherwise applied to the whole page rather than just the SVG. Stylesheets cannot reach into a shadow root, so theme CSS such as `.entry-content svg { fill: red; }` will not apply to those SVGs.
 
-Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered exactly as before and can be styled normally.
+Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered without the shadow root and can be styled by theme stylesheets.
 
 To turn isolation off, at the cost of allowing an SVG's CSS to affect the rest of the page:
 
--- a/includes/blocks/safe-svg/register.php
+++ b/includes/blocks/safe-svg/register.php
@@ -52,6 +52,8 @@ function render_block_callback( $attributes ) {
 	 */
 	$class_name = apply_filters( 'safe_svg_inline_class', 'safe-svg-inline' );
 
+	$has_stylesheet = svg_has_stylesheet( $contents );
+
 	/**
 	 * Whether to isolate this inline SVG inside a shadow root.
 	 *
@@ -65,10 +67,10 @@ function render_block_callback( $attributes ) {
 	 */
 	$use_shadow_dom = (bool) apply_filters(
 		'safe_svg_inline_use_shadow_dom',
-		svg_has_stylesheet( $contents ),
+		$has_stylesheet,
 		$contents,
 		$attributes['imageID'],
-		svg_has_stylesheet( $contents )
+		$has_stylesheet
 	);
 
 	if ( $use_shadow_dom ) {
--- a/includes/blocks/safe-svg/register.php
+++ b/includes/blocks/safe-svg/register.php
@@ -142,6 +142,8 @@ function render_block_callback( $attributes ) {
 /**
  * Check whether an SVG carries its own stylesheet.
  *
+ * Keep in sync with hasStylesheet() in inline-svg.js.
+ *
  * @since x.x.x
  *
  * @param string $contents The SVG contents.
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -72,6 +72,18 @@ add_filter( 'svg_allowed_tags', function ( $tags ) {
 } );
 ```
 
+### Can my theme style an inline SVG?
+
+Mostly, yes. The Inline SVG block renders an SVG that carries its own `<style>` element inside a shadow root, because CSS inside an inline SVG is otherwise applied to the whole page rather than just the SVG. Stylesheets cannot reach into a shadow root, so theme CSS such as `.entry-content svg { fill: red; }` will not apply to those SVGs.
+
+Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered exactly as before and can be styled normally.
```
peterwilsoncc: "exactly as before" doesn't really make sense to a new user.

```suggestion
Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered without the shadow root and can be styled by theme stylesheets.
```

includes/blocks/safe-svg/register.php
```diff
@@ -52,6 +52,27 @@ function render_block_callback( $attributes ) {
...
+	 * @param bool   $use_shadow_dom Whether to isolate the SVG. Defaults to true
+	 *                               when the SVG carries its own stylesheet.
+	 * @param string $contents       The SVG contents.
+	 * @param int    $attachment_id  The ID of the attachment.
+	 */
+	$use_shadow_dom = (bool) apply_filters(
+		'safe_svg_inline_use_shadow_dom',
+		svg_has_stylesheet( $contents ),
+		$contents,
+		$attributes['imageID']
```
peterwilsoncc: Let plugin know whether the SVG has a stylesheet.

```suggestion
		$attributes['imageID'],
		svg_has_stylesheet( $contents )
```

tests/unit/test-safe-svg-block.php
```diff
@@ -0,0 +1,283 @@
...
+				),
+				$attributes
+			)
+		);
+	}
+
+	/**
+	 * Test that SVGs carrying a stylesheet are recognised.
+	 */
+	public function test_svg_has_stylesheet() {
```
peterwilsoncc: Nit: might be nice to use a data provider here so the test completes for each item even if a previous one fails.

tests/unit/test-safe-svg-block.php
```diff
@@ -0,0 +1,283 @@
...
+		$this->assertTrue( Block\svg_has_stylesheet( '<svg><svg:style>a{fill:red}</svg:style></svg>' ) );
+
+		// A self closing element still declares a stylesheet.
+		$this->assertTrue( Block\svg_has_stylesheet( '<svg><style/></svg>' ) );
+	}
+
+	/**
+	 * Test that SVGs without a stylesheet are left alone.
+	 */
+	public function test_svg_has_no_stylesheet() {
```
peterwilsoncc: as above re data provider.

readme.txt
```diff
@@ -69,6 +69,16 @@ They take one argument that must be returned. See below for examples:
         return $tags;
     } );
 
+= Can my theme style an inline SVG? =
+
+Mostly, yes. The Inline SVG block renders an SVG that carries its own `<style>` element inside a shadow root, because CSS inside an inline SVG is otherwise applied to the whole page rather than just the SVG. Stylesheets cannot reach into a shadow root, so theme CSS such as `.entry-content svg { fill: red; }` will not apply to those SVGs.
+
+Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered exactly as before and can be styled normally.
```
peterwilsoncc:
```suggestion
Inherited properties still cross the boundary, so setting `color` on an ancestor and using `currentColor` inside the SVG works, as do CSS custom properties. SVGs that do not contain a `<style>` element are rendered without the shadow root and can be styled by theme stylesheets.
```

peterwilsoncc (commented):
This is testing well and looks good.

I've added a few notes inline but there's nothing major that I have picked up.

readme.txt
dkotter: Resolved in https://github.com/10up/safe-svg/pull/328/commits/2e9fc51577eacb0d37e46870b4f9525f242e871a

README.md
dkotter: Resolved in https://github.com/10up/safe-svg/pull/328/commits/2e9fc51577eacb0d37e46870b4f9525f242e871a

tests/unit/test-safe-svg-block.php
dkotter: Resolved in https://github.com/10up/safe-svg/pull/328/changes/555721f994e0378bd703f1119ed4c786bcf7f203

tests/unit/test-safe-svg-block.php
dkotter: Resolved in https://github.com/10up/safe-svg/pull/328/changes/555721f994e0378bd703f1119ed4c786bcf7f203

includes/blocks/safe-svg/register.php
dkotter: Resolved in https://github.com/10up/safe-svg/pull/328/changes/b74d72192c8c296d83dc0477c6ddca04c9bcc51f and https://github.com/10up/safe-svg/pull/328/changes/555721f994e0378bd703f1119ed4c786bcf7f203

peterwilsoncc (approved):
This looks good to me and is testing well.

Testing notes:

* styled SVGs uploaded on the `develop` branch are rendered in shadow DOM on this branch
* styles within SVGs can no longer effect the site styles (I was testing with `body {background: grey;}`
* SVGs without style elements render without a shadow DOM on this branch
* CSS vars and current color are applied to the SVG

